PR #30, opened 2015-06-10
Fix for legacy publisher + gitignore

MichaelEischer:
The legacy publisher sent an invalid defense area strech.
The gitignore hides the content of the bin and build directory, as done in the svn repo.


joydeep-b:
Good catch, thank you!

PR #34, opened 2015-10-19
Add save settings button, Makefile cleanup

MichaelEischer:
Allows saving the current settings without closing ssl-vision.

The Makefile cleanup ensures that the users home is not deleted when calling "make cleanup_cache" without buildDir being set.


joydeep-b:
Thanks for the updates, Michael.

Issue #33, opened 2015-10-01
USB Camera / OpenCV Support Integration


zaidpirwani:
Hi,

So, we are working with USB cameras, specifically the Playstation 3 EYE Camera, and in order to use that we need USB Camera and OpenCV Support, which we have from an OLD fork of ssl-vision which is here
https://github.com/xuyuan/ssl-vision

Can someone help me, or guide me a bit on how I can integrate the OpenCV support in the current version of ssl-vision, i.e. this repo.

joydeep-b:
Hi Zaid,

There is no official USB support yet. It would be great if you could create
a patch or pull-request with USB camera support. To add USB camera support,
you will need to define a new capture source. For a simple example, see
capturefromfile.h, .cpp in ssl-vision/src/shared/capture/ . Also see how it
is instantiated in ssl-vision/src/app/capture_thread.cpp .

-Joydeep

joydeep-b:
V4L Support added with #35 (Thanks, Eric!).

Issue #28, opened 2015-06-03
How to use my laptop webcam as an input to ssl-vision?


GoogleCodeExporter:
```
How to use my laptop webcam as an input to ssl-vision?


```

Original issue reported on code.google.com by `Fawad.Ah...@gmail.com` on 30 Jan 2014 at 2:01

joydeep-b:
Duplicate: #35 , #36

Issue #11, opened 2015-06-03
USB-Cameras


GoogleCodeExporter:
```
I would like to know if you work on configrung the SSL_Vision software so that 
it works with USB-cameras.

```

Original issue reported on code.google.com by `unice...@googlemail.com` on 6 Aug 2010 at 10:28

joydeep-b:
Added V4L support in #35

PR #35, opened 2016-02-05
Feature/v4l usb camera

ezavesky:
* adding early usb camera (via v4l) support to ssl-vision
   * Note: currently lacks ability to set alternate frame resolution and capture rate
* attempts to satisfy some open issues like #33 and #28, maybe even issue #11 (but it does not include OpenCV hooks)
* also sneaks in a few updates to the main README.md file to reflect new usb functionality


src/app/capture_thread.cpp
```diff
@@ -184,39 +190,40 @@ void CaptureThread::run() {
...
               capture_mutex.unlock();
-            }
-            stack_mutex.lock();
-            stack->updateTimingStatistics();
-            stack_mutex.unlock();
-          }
-          capture_mutex.lock();
-          if ((capture != 0) && (capture->isCapturing())) {
-            capture->releaseFrame();
           }
```
joydeep-b: Should have some error statement else?
i.e. if bSuccess == false

src/app/gui/glLUTwidget.cpp
```diff
@@ -781,7 +783,8 @@ void GLLUTWidget::glDrawSlice(Slice * s) {
     }
     glEnable(GL_COLOR_LOGIC_OP);
     //glLogicOp(GL_COPY_INVERTED);
-    glLogicOp(GL_XOR);
+    //glLogicOp(GL_XOR);
+    glLogicOp(GL_EQUIV);                //fix drawing inverted color (2/2/16)
```
joydeep-b: What is this for?

src/shared/CMakeLists.txt.inc
```diff
@@ -12,10 +12,11 @@ include_directories(${shared_dir}/vartypes/enhanced)
 include_directories(${shared_dir}/vartypes/gui)
 include_directories(${shared_dir}/vartypes/xml)
 
-include_directories("/usr/include/eigen2")
+# include_directories("/usr/include/eigen2")  # update for eigen3 at toplevel 01/16
```
joydeep-b: Is this comment correct? If not, please remove.

src/shared/capture/capturev4l.h
```diff
@@ -0,0 +1,350 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string>
+#include "VarTypes.h"
+#include <linux/videodev2.h>
+#include <sys/poll.h>
+
+#include <map>
+
+//#include "conversions.h"
```
joydeep-b: Please remove.

src/app/gui/glLUTwidget.cpp
ezavesky: not sure why the original inversion got commented out, but this line restores the ability to see the LUT marked up with a sample drawn from the active image (now they are visualized with small white 'x's as I think was originally intended)

src/app/capture_thread.cpp
ezavesky: Unfortunately, the auto re-indent caused some mess here, but the main reason is that on a good frame we'll proceed for the whole frame/update but we skip the process on a bad frame.  At the time there was some race condition (e.g. when starting with '-s') that caused a crash with some slow-to-start drivers (like v4l).

src/shared/util/rawimage.cpp
```diff
@@ -150,20 +150,13 @@ void RawImage::clear()
...
-    case COLOR_MONO16:
-    return pixelCount*2;
+    case COLOR_RGB8: return pixelCount*3;
+    case COLOR_RGBA8: return pixelCount*4;
+    case COLOR_YUV444: return pixelCount*3;
+    case COLOR_YUV422_UYVY: return pixelCount*2;
+    case COLOR_YUV411: return pixelCount*3/2;
+    case COLOR_MONO8: return pixelCount;
+    case COLOR_MONO16: return pixelCount*2;
     default:
```
joydeep-b: If changing the previous cases, please change this to be consistent, as well.

src/shared/util/rawimage.h
```diff
@@ -38,7 +38,7 @@
 */
 class RawImage : public ImageInterface
 {
-  protected:
+protected:
```
joydeep-b: Please revert.

src/shared/util/rawimage.h
```diff
@@ -54,7 +54,7 @@ class RawImage : public ImageInterface
   /// capture timestamp of the image
   double   time;
 
-  public:
+public:
```
joydeep-b: Please revert.

src/app/capture_thread.cpp
```diff
@@ -36,9 +36,11 @@ CaptureThread::CaptureThread(int cam_id)
   control->addChild( (VarType*) (captureModule= new VarStringEnum("Capture Module","DC 1394")));
   captureModule->addFlags(VARTYPE_FLAG_NOLOAD_ENUM_CHILDREN);
   captureModule->addItem("DC 1394");
+  captureModule->addItem("V4L (USB)");
```
joydeep-b: Please call this "Video 4 Linux", to be precise. PtGrey and other manufacturers have their own custom USB interface.

src/shared/CMakeLists.txt.inc
ezavesky: Yes, on some lesser devices (e.g. raspberry-pi ubuntu) there was some difficulty getting the eigen2 library, so the cmake dependency and single line of code were updated. (see 816042cf8d9c5f95f77a82c589c2b12c8772a62c)

joydeep-b:
Thanks for the pull-request, Eric. I reviewed it, and it looks pretty good, but I had some comments (above). Could you please update based on the comments? Thanks.

src/shared/capture/capturev4l.h
joydeep-b: (I meant the commented out line)

ezavesky:
Heh, looks like some confusion with the first commit, so we'll aim for something more graceful next time.  I was going to pull request from an isolated branch so that other changes could be pushed, but I can resolve this first (and keep this history) if you prefer.

src/shared/capture/capturev4l.h
ezavesky: Not sure what is incorrect about this commit.  Was it the commented-out "conversions.h" include?

joydeep-b:
Can you edit the code based on the comments inline here? Then that'll ensure that the requested edits happen before pulling these changes.

src/shared/util/rawimage.cpp
ezavesky: No functional change, just over-zealous readability.

src/shared/CMakeLists.txt.inc
joydeep-b: I see. Okay, given that the primary platform (Ubuntu LTS) has Eigen3, then we could just remove this commented out line entirely.

src/app/capture_thread.cpp
ezavesky: Fixed in ad663fcf87c382778415e1955a3ad7f265a910f1

src/shared/util/rawimage.h
ezavesky: Fixed in ad663fcf87c382778415e1955a3ad7f265a910f1

src/shared/util/rawimage.h
ezavesky: Fixed in ad663fcf87c382778415e1955a3ad7f265a910f1

src/shared/CMakeLists.txt.inc
ezavesky: Updated in change 89e70a6e2b6cebfd6f133d7e1e897600c2e8a85e (currently in testing)

ezavesky:
@joydeep-b Okay, all issues have been addressed, please pull at your convenience.   I have another fix/feature for the generator that I think would be useful but can't push because this PR is open on the master branch.

joydeep-b:
LGTM.

MichaelEischer:
Nice to see some new features for SSL-Vision.
I've noticed some comments containing a date and reason why something was changed. For example "//fix drawing inverted color (2/2/16)". along with leaving some old dead code as comments in place. This kind of notes is, however, best managed by the source code management tools. So in my opinion this should be avoid unless there is a good reason for it. `git blame` `git log` and `git revert` (just in case the change wasn't that good after all) would be my preferred tools to replace those comments. (Sorry for nit-picking)
P.S. GitHub managed to send me 28 emails for this issue... ^^

PR #37, opened 2016-02-05
- remove lingering debug messages, fix GUI race condition on close

ezavesky:
Missed these after last debug session, but makes output cleaner for exit and if you don't care about V4L cameras.

### DIFF
--- a/src/shared/capture/capturev4l.cpp
+++ b/src/shared/capture/capturev4l.cpp
@@ -1047,10 +1047,8 @@ void CaptureV4L::cleanup()
 #endif
     
     //TODO: cleanup/free any memory buffers.
-    /*if (camera_instance && !camera_instance->stopStreaming()) {
-        fprintf(stderr,"CaptureV4L Error: Unable to stop streaming, was camera started?\n");
-    }
-    */
+    if (camera_instance && is_capturing)
+        camera_instance->stopStreaming();
     is_capturing=false;
 #ifndef VDATA_NO_QT
     mutex.unlock();
--- a/src/shared/capture/capturev4l.cpp
+++ b/src/shared/capture/capturev4l.cpp
@@ -226,13 +226,9 @@ int GlobalV4Linstance::xioctl(int fd,int request,void *data,
     }
     
     // report error
-    if(ret != 0){
-        if(!error_str) error_str = "";
-        //        AnsiColor::SetFgColor(stderr,AnsiColor::Red);
-        //        AnsiColor::Bold(stderr);
+    if(ret != 0 && error_str) {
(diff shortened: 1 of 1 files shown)


src/shared/capture/capturev4l.cpp
```diff
@@ -1050,16 +1046,11 @@ void CaptureV4L::cleanup()
...
-//    if (camera_instance!=0) {
-//        dc1394_capture_stop(camera);
-//        dc1394_video_set_transmission(camera,DC1394_OFF);
-//        dc1394_camera_free(camera);
-//    }
-//    camera=0;
     //TODO: cleanup/free any memory buffers.
-    if (camera_instance && !camera_instance->stopStreaming()) {
+    /*if (camera_instance && !camera_instance->stopStreaming()) {
         fprintf(stderr,"CaptureV4L Error: Unable to stop streaming, was camera started?\n");
```
joydeep-b: Can these commented out lines be removed now?

ezavesky:
@joydeep-b I like the way you think.  This is a safer and cleaner way to handle it.

src/shared/capture/capturev4l.cpp
```diff
@@ -581,7 +577,8 @@ CaptureV4L::CaptureV4L(VarList * _settings,int default_camera_id) : CaptureInter
     }
     
     if (cam_id > cam_list[cam_count-1]) {
-        fprintf(stderr,"CaptureV4L Error: no camera found with index %d. Max index is: %d\n",cam_id,cam_list[cam_count-1]);
+        //fprintf(stderr,"CaptureV4L Error: no camera found with index %d. Max index is: %d\n",cam_id,cam_list[cam_count-1]);   /// 2/5/16 --- shh, please be quiet
```
joydeep-b: I'm a bit uncomfortable about silently gobbling up an error message like this, which might actually help the user. The DC1394 driver explicitly prints an error when asked to open a camera Idx > num_cameras - 1. 
How about a printonce-like functionality as a compromise?
{
  static bool printed_once = false;
  if (!printed_once) {
    fprintf(etderr, "Error... This message will print only once.");
    printed_once = true;
  }
}

ezavesky: In reality, this doesn't happen very often (only at startup when it attempts to enumerate cameras) but I didn't want to confuse users that start seeing responses about V4L cameras when they're used to the dc1394 or others.  A one-time message is acceptable for this -- but it will actually get printed 4 times -- once for each camera.

joydeep-b: Okay, sounds good. We can revisit if necessary after more people have tested,

PR #39, opened 2016-02-05
Feature/generate team

ezavesky:
* add a new mode to inject team codes into generator for easier LUT training (see image); towards easier LUT sync across cameras (requested in [legacy TODO](https://github.com/RoboCup-SSL/ssl-vision/blob/wiki/TODO.md))
* add a method to search for nodes in VarType tree (allows settings from one module to be pulled into another -- like teams into capture_generator)
* optimize generator to cache frames instead of always redrawing; doubled frame rate throughput

160205_generate_team

### DIFF
--- a/src/app/gui/glwidget.cpp
+++ b/src/app/gui/glwidget.cpp
@@ -326,7 +326,7 @@ void GLWidget::callZoomFit() {
 
 void GLWidget::callHelp() {
   QMessageBox::information ( this,"Keyboard and Mouse shortcuts",
-                             "Left-Click: Pick RGB Color\nShift-Modifier: Erase pixel from LUT\nCtrl-Modifier: Add pixel to LUT\nArrow-Keys (or Right-Drag Mouse): Panning\nPage Up/Down (or Mousewheel): Zooming\nHome: Reset Zoom to 100%\nSpace: Reset Zoom to Best Fit\nCtrl-f: Toggle Fullscreen\nCtrl-w: Toggle Detach Window" );
+                             "Left-Click: Pick RGB Color\nShift-Modifier: Erase pixel from LUT\nShift-Ctrl-Modifier: Add pixel to LUT\nArrow-Keys (or Right-Drag Mouse): Panning\nPage Up/Down (or Mousewheel): Zooming\nHome: Reset Zoom to 100%\nSpace: Reset Zoom to Best Fit\nCtrl-f: Toggle Fullscreen\nCtrl-w: Toggle Detach Window" );
 }
 
 void GLWidget::keyPressEvent ( QKeyEvent * event ) {
--- a/src/shared/capture/capture_generator.cpp
+++ b/src/shared/capture/capture_generator.cpp
@@ -17,11 +17,16 @@
   \brief   C++ Implementation: CaptureGenerator
   \author  Stefan Zickler, (C) 2009
 */
+// 2/4/16 - added ability to load team patterns from file (Zavesky)
 //========================================================================
 
 #include "capture_generator.h"
 #include "conversions.h"
 
+#include "cmpattern_team.h"         //added to load team files directly
+#include "cmpattern_teamdetector.h"
+
+#define TEAM_IMAGE_PLACEHOLDER      "(team images - reset bus to populate)"
 
 #ifndef VDATA_NO_QT
 CaptureGenerator::CaptureGenerator ( VarList * _settings, QObject * parent ) : QObject ( parent ), CaptureInterface ( _settings )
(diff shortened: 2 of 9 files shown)


joydeep-b:
The old pattern definitions are currently deprecated, and no longer used by the RoboCup SSL. Also, the generator is not intended to be used for color calibration, since color calibration requires the actual captured images, not synthetic images. 

The other two features could potentially be useful: could you please refactor them into separate commits? In general, we try to separate functionality across commits so that features, bugs, and changes can be localized to one segment of the codebase per commit.

Thanks.

ezavesky:
Thanks for update about patterns, are there newer patterns available for testing?  
I believe this feature to be critical because otherwise there is no link in the LUT to the actual patterns.  For example, for the longest time, we were training on live images under a certain color name but there was no link to the static colors if you started from an empty LUT -- which I advise because there is a lot of noise for a particular camera.

Addressing your second request, what two functions are you referring to?  To permit search through the xml/vartype structure, a necessary vartype was introduced to allow hierarchy traversal.  Happy to oblige, but not sure how to get static image information into a LUT if not this way.

joydeep-b:
The Standard2010.png pattern is the only pattern used now.
I'm not sure I understand the link between the LUT to patterns. The LUT provides a mapping from YUV space to color labels, but color labels are independent of patterns. The mapping however is specific to the camera settings and lighting conditions, and do not generalize.

ezavesky:
Sampling from a static file is still required because a blank/new LUT will not have any information about color that _should_ exist in a pattern.  Specifically, this pattern resulted in no detections while training with a new camera.
1. Start ssl-vision with a clean slate
2. Start the feed for a new camera (the device/capture does not specifically matter)
3. Mark-up the LUT using correct color assignment for a camera (e.g. label the blue, green, yellow) with the signal received and present within the vision app
4. No robots will ever be detected because there is no correspondence between the new live LUT markup and the static pattern images.  The viable remedy (which drove this change) was the introduction of static images into the app so that LUT color entries (green, blue, etc) would map to the pattern image itself.

joydeep-b:
Steps 1-3, along with geometry calibration, should be sufficient to detect
robot patterns. It is the workflow we use when setting up ssl-vision on a
new field, or computer.
The pattern images themselves are not important beyond setting the
parameters for blob and pattern geometry. The pattern images are not used
for anything related to the LUT.

If steps 1-3, along with geometry calibration (the geometry calibration is
crucial: the pattern detector relies on it) did not work to detect robots
in your case, then we need to investigate that further on a separate
debugging thread. There should not be any need for injecting pattern images
into the generator.

ezavesky:
FYI: A new branch has been created at [Emoters:feature/vartype_parents](https://github.com/Emoters/ssl-vision/commits/feature/vartype_parents) for the vartype changes required by other PRs (although they might need to be hand-merged).  I would prefer to conclude this discussion before going that route though.

The remaining confusion comes from this phrase in your last comment...
> The pattern images themselves are not important beyond setting the parameters for blob and pattern geometry. The pattern images are not used for anything related to the LUT.

This is the requirement that is not satisfied without an initial inspection of the pattern image _with_ some form of LUT color annotation.  While you may be correct that the LUT doesn't care about the pattern image after calibration, there must be some step to bootstrap the LUT (specifically the color/name indices which determine region geometry) before any detection occurs.  As a newcomer it's possible that I'm mistaken, but this is a required action that does not correlate to any faulty image or other detection code.

joydeep-b:
The color names / indices are already hardcoded in. See:
https://github.com/RoboCup-SSL/ssl-vision/blob/master/src/shared/cmpattern/cmpattern_teamdetector.h

joydeep-b:
Also, the channel IDs are initialized when the SSL Vision stack is created:
https://github.com/RoboCup-SSL/ssl-vision/blob/master/src/app/stacks/stack_robocup_ssl.cpp#L47
https://github.com/RoboCup-SSL/ssl-vision/blob/master/src/shared/util/lut3d.h#L349

ezavesky:
Yes, while the colors are hard coded, I was looking for correspondence between a color and its placement in a pattern.  This appears to occur in [team pattern initialization](https://github.com/RoboCup-SSL/ssl-vision/blob/master/src/shared/cmpattern/cmpattern_teamdetector.cpp#L147) where the hard-coded colors are quantized into a new LUT and region placement and size are learned as a model.  The implications here are that these colors are very critically cemented as what must be used for bot detection.  I have created issue #46 as a bookmark (and possible tombstone) for this discussion.

I will refactor this PR but keep the pattern injector change active on our personal fork for later use.

PR #40, opened 2016-02-07
Hotfix/live lut update

ezavesky:
* update to sync LUT when colors from live video are added (ctrl+shift+click) or removed (shift+click)
   * previously, LUT was updated, but derived tables were not, which means no processing update
* one minor tweak for printing errors in v4l module (instead of hiding all responses)
* update URL from old google code pointer to github


joydeep-b:
Looks good, thanks for the update.

PR #41, opened 2016-02-08
Feature/lut copy

ezavesky:
*  add button in "YUV Calibrator" to copy LUT from other cameras (solving [legacy request](https://github.com/RoboCup-SSL/ssl-vision/blob/wiki/TODO.md) to allow easy LUT channel sync)
   * **NOTE** builds on some vartype changes in PR #39 
   * a space to limit which color is copied, but not currently implemented (let's gauge the interest first)
* update vartype to allow access to parent node

160207_lut_copy

### DIFF
--- a/src/app/gui/glLUTwidget.cpp
+++ b/src/app/gui/glLUTwidget.cpp
@@ -897,9 +897,11 @@ void GLLUTWidget::setLUT(LUT3D * lut) {
 }
 
 bool GLLUTWidget::copyLUT(void *pDataLUT, int size_copy, int color_index) {
-    editStore();                            //edit snapshot
+    //edit snapshot
+    editStore();                            
     m.lock();
-    bool bSuccess = _lut->copyLUT(pDataLUT, size_copy, color_index);  //perform actual copy
+    //perform actual copy
+    bool bSuccess = _lut->copyLUT(pDataLUT, size_copy, color_index);  
     m.unlock();
     if (bSuccess) rebuildAndRedraw();
     return bSuccess;
--- a/src/app/gui/glLUTwidget.cpp
+++ b/src/app/gui/glLUTwidget.cpp
@@ -896,6 +896,15 @@ void GLLUTWidget::setLUT(LUT3D * lut) {
   m.unlock();
 }
 
+bool GLLUTWidget::copyLUT(void *pDataLUT, int size_copy, int color_index) {
+    editStore();                            //edit snapshot
+    m.lock();
+    bool bSuccess = _lut->copyLUT(pDataLUT, size_copy, color_index);  //perform actual copy
+    m.unlock();
+    if (bSuccess) rebuildAndRedraw();
+    return bSuccess;
+}
+
 void GLLUTWidget::init() {
   if (_lut!=0) {
     int z=_lut->getSizeX();
--- a/src/app/plugins/plugin_colorcalib.cpp
+++ b/src/app/plugins/plugin_colorcalib.cpp
@@ -31,7 +31,7 @@ PluginColorCalibration::PluginColorCalibration(FrameBuffer * _buffer, YUVLUT * _
     settings=new VarList("YUV Calibrator");
     continuing_undo = false;
     
-    copy_LUT_trigger = new VarTrigger("Copy LUT", "Copy LUT!");
+    copy_LUT_trigger = new VarTrigger("Copy LUT", "Copy LUT");
     settings->addChild(copy_LUT_trigger);
     connect(copy_LUT_trigger,SIGNAL(signalTriggered()),this,SLOT(slotCopyLUT()));
     settings->addChild( v_lut_sources = new VarStringEnum("LUT Source", LUT_COPY_PLACEHOLDER));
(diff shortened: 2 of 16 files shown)


src/app/gui/glLUTwidget.cpp
```diff
@@ -896,6 +896,15 @@ void GLLUTWidget::setLUT(LUT3D * lut) {
   m.unlock();
 }
 
+bool GLLUTWidget::copyLUT(void *pDataLUT, int size_copy, int color_index) {
+    editStore();                            //edit snapshot
```
joydeep-b: Please put comments on the preceding line.

src/app/gui/glLUTwidget.cpp
```diff
@@ -896,6 +896,15 @@ void GLLUTWidget::setLUT(LUT3D * lut) {
...
+    m.lock();
+    bool bSuccess = _lut->copyLUT(pDataLUT, size_copy, color_index);  //perform actual copy
```
joydeep-b: Ditto.

src/app/plugins/plugin_colorcalib.cpp
```diff
@@ -21,13 +21,23 @@
 #include "plugin_colorcalib.h"
 #include <QStackedWidget>
 
+#define LUT_COPY_PLACEHOLDER    "(click copy to refresh)"
```
joydeep-b: Please do not use #defines unless absolutely necessary.
This functionality can be reproduced with the use of a static const string variable in the scope of the function using it.

src/app/plugins/plugin_colorcalib.cpp
```diff
@@ -21,13 +21,23 @@
...
-  lutw=0;
-  settings=new VarList("YUV Calibrator");
-  continuing_undo = false;
+    mode=_mode;
+    lut=_lut;
+    lutw=0;
+    settings=new VarList("YUV Calibrator");
+    continuing_undo = false;
+    
+    copy_LUT_trigger = new VarTrigger("Copy LUT", "Copy LUT!");
```
joydeep-b: Please remove the "!".

src/app/plugins/plugin_colorcalib.cpp
```diff
@@ -140,3 +153,67 @@ void PluginColorCalibration::mouseReleaseEvent ( QMouseEvent * event, pixelloc l
...
+            v_lut_colors->selectIndex(0);
+        }
+    }
+    else {              //user has selected a camera
+        bool bSuccess = true;
+        VarBlob *pSource = NULL;
+        int color_index = -1;       //default -1 is ALL colors; TODO be from var 'v_lut_colors'
+        std::vector<VarType *> vectTarget = settings->findRelatives(v_lut_sources->getString(), true);
+        if (vectTarget.size()!=1) {
+            fprintf(stderr, "LUT3D: Found too many cameara nodes (%d) for camera '%s', aborting copy.\n",
```
joydeep-b: Typo: cameara-> camera

src/app/plugins/plugin_colorcalib.cpp
```diff
@@ -140,3 +153,67 @@ void PluginColorCalibration::mouseReleaseEvent ( QMouseEvent * event, pixelloc l
...
+        }
+        if (bSuccess && vectTarget.size()!=1) {
+            fprintf(stderr, "LUT3D: Found too many LUT nodes (%d) for camera '%s', aborting copy.\n",
+                    static_cast<int>(vectTarget.size()), v_lut_sources->getString().c_str());
+            bSuccess = false;
+        }
+        else {
+            pSource = reinterpret_cast<VarBlob*>(vectTarget[0]);
+            
+            //WARNING: Makes assumption that all LUT blobs have the same properties
```
joydeep-b: This sounds dangerous. How about doing a deep copy instead? It'll take up a bit more (kilobytes?) of memory, but prevent future memory errors.

src/shared/capture/capture_generator.cpp
```diff
@@ -17,11 +17,17 @@
   \brief   C++ Implementation: CaptureGenerator
   \author  Stefan Zickler, (C) 2009
 */
+// 2/4/16 - added ability to load team patterns from file (Zavesky)
```
joydeep-b: We do not need this feature - please see my comment for #39

src/shared/vartypes/primitives/VarStringEnum.h
```diff
@@ -119,15 +119,15 @@ namespace VarTypes {
   
     /// get the string of item at a given index
```
joydeep-b: Please revert whitespace changes.

src/app/plugins/plugin_colorcalib.cpp
ezavesky: Correct, but this is used in several places, which requires the exact string and is likely safer for a single modification point.

joydeep-b:
Thanks for working on a copy LUT feature: it would be very useful.
I have added some comments to the code, could you please update?
There seem to be three independent feature changes here, could you please refactor them into separate commits? The pattern in generator injector is not something that we'd want on the main branch (see my comment on #39

src/app/plugins/plugin_colorcalib.cpp
ezavesky: A relic of copying the code for saving the XML file, it can be updated.

src/app/plugins/plugin_colorcalib.cpp
ezavesky: Unfortunately, the way the VarTypes are used to deeply copy and sequester data, there is no great way to propagate a modification through the rest of the application (through to processing, data storage, LUT).  Admittedly the copy is dangerous, but it is size-checked in the LUT class itself and there aren't any other great ways that I could find in the near-term.

Issue #43, opened 2016-02-08
request for installable packge that could be seeded into deb/rpm repos


ezavesky:
Request for a packaged binary version that correctly enumerates (and installs) required libraries.  With the availability of modern [CI services](https://travis-ci.org/) and other resources, it would be great if ssl-vision could be built and distributed through a common deb package.  There appear to be some [control files](https://github.com/RoboCup-SSL/ssl-vision/tree/master/debian) in the tree already, but it's not clear how and when they are used.  Below are proposed acceptance criterion for this issue.
- [ ] a packaged deb file for at least Ubuntu 12.X and Ubuntu 14.X
  - both of these Ubuntu images are available as CI environments within travis
  - **proposed solution**: utilization of the [CPack functionality](https://cmake.org/Wiki/CMake:Packaging_With_CPack) available in CMake (current build driver)
- [x] a stable build indicator (and integrator) for new development PR's and commits so that the burden on the project maintainer is reduced (and acceptance time is minimized)
  - **proposed solution**: Introduction of a [.travis.yml](https://docs.travis-ci.com/user/getting-started/) file and update of CMake utilities to use the [packaging utilities](https://cmake.org/Wiki/CMake:Packaging_With_CPack) already available
  - typically with modern CI's, this functionality will come for free, but it does require a few github authorizations from the project manager

ezavesky:
Not wanting to cloud the issue itself, efforts to bring ssl-vision into Travis are underway at [a forked integration branch](https://github.com/Emoters/ssl-vision/tree/integration).  Before bringing these changes to ssl-vision, the excessive/experimental commits (is there an easier way with a network-distributed service?) will be pruned.

g3force:
I tried introducing such installable packages in other applications in the past and my experience is: It is not worse the effort. Some reasons:

 - You will never have a package for all distributions
 - You have to refactor the way, configuration files are stored, because there is no unique working directory anymore. As you still want to have the possibility to run ssl-vision from a Repo, you would need a way to conditionally place config files.
 - it creates even more distance to the code, discouraging people to contribute

ezavesky:
Sure, solving everyone's requirements simultaneously may not be feasible.  The main objectives of a CI process were (a) automatic building of a reference that would immediately throw bugs (and offer a route for a test harness) and (b) let beginners get started with some package if they wanted to.  Towards your objections on (b) was the ability to jump in with a VM and get something running.  

[...]

PR #45, opened 2016-02-09
Feature/travis ci build

ezavesky:
- add [travis-ci](https://travis-ci.org/) support for build structure on ubuntu 12.x (precise) and 14.x (trusty)
  - _NOTE_: requires create of new application authorization for this repo as well
  - minor syntax fix for recently added usb capture class
  - _NOTE 2_: adding the cmake file for Eigen3 is not desired, but the package is not installing consistently across different ubuntu distributions, so this is the best workable solution
- partial solution to issue #43 because new builds and pull requests will be vetted automatically

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+language: cpp
+os: linux
+sudo: required
+env: BUILD_TYPE=Release
+
+compiler:
+  - gcc
+
+before_install:
+  - sudo apt-get update -qq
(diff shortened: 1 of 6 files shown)


g3force:
I created a new PR with only the travis file: https://github.com/RoboCup-SSL/ssl-vision/pull/83
I think, the rest is outdated or unrelated.
PR can be closed.

ezavesky:
Thanks, @g3force.  It has been a while and I can't get back to that system for testing.  Having auto-builds would still be great though so happy to pass the torch.

PR #49, opened 2016-02-10
Fix LUT single slice view.

MichaelEischer:
This commit reverts 8bea8d5f151c8af0c894d3310000b181da8e98e2 and cleans up the existing code.

The sampler layer is either transparent black or solid white. The transparent black parts are ignored when drawing.
Using xor with solid white is identical to inverting a pixel. GL_EQUIV is the exact opposite of xor, that is solid white causes no pixel changes and transparent black apparently too.
The views other than single slice view seem to always draw the sample marks due to scaling artifacts (switch to cube mode and move the camera near and far away from the cube).

Using GL_EQUIV break the single slice view, which is probably the most important one for doing color calibration.

### DIFF
--- a/src/app/gui/glLUTwidget.cpp
+++ b/src/app/gui/glLUTwidget.cpp
@@ -782,11 +782,12 @@ void GLLUTWidget::glDrawSlice(Slice * s) {
         glVertex3f(1.0 ,0.0  ,z+z_offset);
       glEnd();
     }
-    glEnable(GL_COLOR_LOGIC_OP);
-    //glLogicOp(GL_COPY_INVERTED);
-    //glLogicOp(GL_XOR);
-    glLogicOp(GL_EQUIV);                //fix drawing inverted color (2/2/16)
     if (s->sampler->bind()) {
+      glEnable(GL_COLOR_LOGIC_OP);
+      // invert color+alpha below the sampling marks to ensure visibility
+      // testing this operation should be done when only a single slice is displayed
+      // the other views seem to always draw the sample marks due to scaling artifacts
+      glLogicOp(GL_XOR);
       glBegin(GL_QUADS);
         glTexCoord2f(0.0,0.0);
         glVertex3f(0.0  ,0.0,z+z_offset*2);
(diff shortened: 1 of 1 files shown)


ezavesky:
For me this change breaks two things that were recently fixed, can you confirm them first?
 1.  The 'x' indicators for where an image's presence in the LUT is broken (typing 'i' with on a live image).  It longer works in single, multiple, or box-view.
 2. When drawing on live mode (click+ctrl+shift from an image or just normal click on LUT) the inverse/negative color is drawn in the LUT. For example, when I draw 'field green' a dull pink color is drawn on the LUT.  When I draw 'pink' a green is drawn.

MichaelEischer:
Can you comment out everything between glEnable(GL_COLOR_LOGIC_OP) and glDisable(GL_COLOR_LOGIC_OP)? That way the x indicators will no longer work, while the drawn colors on the LUT must be correct.

In 2. you are referring to the single slice view? The LUT code has been unchanged for 6 years and I'm pretty sure I've never seen the inverted color problem you are describing (its not happening on my machine, I can test it in our lab on monday). The only thing that should be drawn using inverted colors is the 'x' indicator itself.

Cube mode with inverted marks
I've taken that screenshot using my branch. If Gl_EQUIV is used the x-indicators disappear once you zoom in far enough, which I think is an indication that these were just drawing artifacts.

ezavesky:
Okay, your screenshot helped a lot to understand the difference in performance.  Attached is a screenshot from the ssl-vision master branch.  I also compared performance on a different machine, and that's what identified the problem.  Most development has been on a VM (VMWare) on MacOS, which apparently has trouble implementing the XOR logic.  I confirmed that the same code on a native Ubuntu instance did perform as you indicated above.

Given this performance, I would suggest we go for more compatibility (without the XOR), but maybe this is where @joydeep-b can make a call as the repo manager.

160212_lut_test

MichaelEischer:
@joydeep-b could you merge this pull request as a fix for the regression?
Going for more compatibility should be done in a separate patch. (I would actually prefer a fixed VMWare graphics driver).

joydeep-b:
Merged, thanks for the revert.

PR #51, opened 2016-02-10
Load field defaults if no lines and arcs are configured

MichaelEischer:
When ssl-vision is started for the first time, no lines and arcs are added. Getting a working configuration requires clicking the "Reset SSL 2014"-button. This patch automatically loads the defaults if no lines and arcs are set.

A different implementation would be to set the defaults if no settings file exists. The implementation in my patch will, however, also work with existing configurations that lack field lines.


joydeep-b:
Thanks for this patch, Michael. It will eliminate one common pitfall for first-time users. :)

PR #52, opened 2016-02-10
Source compatibility with Qt 5

### LINKED ISSUE #24: Graphical Client Issue
```
Hello everyone This is Antonio from the robobulls team I am having a bit of an 
issue. I am transferring our stuff from windows to Linux we managed to get 
everything set up vision system works and so does the graphical client.
I am trying to enhance the graphical client by adding some extra buttons but to 
even consider tackling that I first need to open and compile the graphical 
client in Qt I believe I found the correct source and after a couple of changes 
I got through a few minor details, the thing holding me back now is an error I 
do not know how to fix or get around I tried search engines but there was 
literally no results for my type of error I am hoping someone may have familiar 
knowledge of the situation or Qt the error is CentralWindow.cpp:119: error: 
'UnicodeUTF8' is not a member of 'QApplication' I have looked at the encoder 
and it is set to UTF8 that was one of the suggestions I found online aside from 
that I am stumped hope someone can assist. 

I am using QtCreator 2.62 and Qt 5.01
I have attached the file in question as well it is all files from the GClient 
provided in the SSL download package  
```

Original issue reported on code.google.com by `RoboBu...@gmail.com` on 27 Mar 2013 at 6:01

Attachments:
- [CentralWindow.cpp](https://storage.googleapis.com/google-code-attachments/ssl-vision/issue-24/comment-0/CentralWindow.cpp)

MichaelEischer:
These changes should fix most incompatibilities with Qt 5 while staying compatible with Qt 4.
For a switch to Qt 5 an additional modification of the CMakeLists.txt is needed.

Fixes #24.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,9 @@ if(USE_QT5)
 	if(POLICY CMP0043) # compatibility with CMake 3.0.1
 	    cmake_policy(SET CMP0043 OLD)
 	endif()
+	if(POLICY CMP0071) # compatibility with CMake 3.10.0
+	    cmake_policy(SET CMP0071 OLD)
+	endif()
 else()
 	cmake_minimum_required(VERSION 2.6)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,16 @@
 project(ssl-vision)
-cmake_minimum_required(VERSION 2.6)
+set(USE_QT5 FALSE CACHE BOOL "Use Qt5 instead of Qt4")
+if(USE_QT5)
+	cmake_minimum_required(VERSION 2.8.9)
+	if(POLICY CMP0020)
+	    cmake_policy(SET CMP0020 NEW) # remove if CMake >= 2.8.11 required
+	endif()
+	if(POLICY CMP0043) # compatibility with CMake 3.0.1
+	    cmake_policy(SET CMP0043 OLD)
+	endif()
+else()
+	cmake_minimum_required(VERSION 2.6)
+endif()
 
 #defines
 add_definitions(-D UNIX)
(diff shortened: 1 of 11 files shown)


ezavesky:
Can you markup or add a conditional compile to the CMake file for someone to switch easily in this PR?  Unless it's agreed to move on to QT5 for all, this will allow an easier switch to it for those who need additional functionality.

MichaelEischer:
The intention of this pull request was to fix compile errors with Qt5 while staying fully compatible with Qt4. However, you're right that it's not too much use without an update CMake file.

The only downside in having a Qt4/Qt5-CMake file is that it's somewhat ugly.

MichaelEischer:
Rebased the branch onto the latest master branch. The third commit was replaced with two fixes for qt warnings while processing the qrc and ui files.

g3force:
@MichaelEischer could you update your branch to resolve the merge request?

Just for my understanding: The issues regarding QT5 compatibility should already be fixed in master, shouldn't they? I think I only have QT5 installed on my system.

MichaelEischer:
ssl-vision on master can definitely only compile using Qt4. See for example CMakeLists.txt#L15 , all my source compatibility changes are also not applied.
The merge commit allows using Qt5 by setting a configuration option. Right now the default is still Qt4.

g3force:
you are right, the qt4 package in Fedora is called "qt-devel" while for qt5 it is called "qt5-devel" -.-

g3force:
When I compile with QT5, I get a warning:
```
Policy CMP0071 is not set: Let AUTOMOC and AUTOUIC process GENERATED files.
  Run "cmake --help-policy CMP0071" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
```

MichaelEischer:
Sorry for the warning, the cmake version on Ubuntu 16.04 didn't complain.

asctec-andre:
I can confirm that this still runs fine with Qt4 on Ubuntu 14.04. When compiling with Qt5 it crashes immediately with a segfault. I assume the Qt5 version on 14.04 is just too old :/

MichaelEischer:
My guess would be that there's some bad interaction between the opengl- and the normal Qt-Widgets. The early Qt5 versions had several problems in that regard.
I could add a minimum version for the Qt libraries. Qt5 >= 5.5.1 (used by Ubuntu 16.04) definitely work, Qt5 <= 5.2.1 (Ubuntu 14.04) probably not. This would leave Qt 5.3 and 5.4 for testing, but I'm not sure whether that's worth the trouble.

g3force:
I agree we should not spent too much effort in the QT5 version issue. We still primary use QT4 and multiple people have tested in successfully, so I will merge this PR now.

PR #53, opened 2016-02-11
Fix compiler warnings and default to -Wall

MichaelEischer:
This fixes all compiler warnings using g++ 5.2.1 (Ubuntu 15.10). Compiling should now also be possible using clang, but that results in a strange linker error for me (probably an incompatibility between g++ 5 and clang related to std::string).

Most changes are unsigned vs signed int, unused variables.

Special ones:
- Allocate marker array in _cmpattern_teamdetector.cpp_ using `new[]`. This allows compilation using clang
- Remove the TeamDetectorInterface from _cmpattern_teamdetector.h_. The interface is not used and defines an update function that differs from the one that's actually used.
- Used isnan from cmath in _glcamera.h_ .
- Fix operators in _gvector.h_ : `+=` would expand to `x = x += p.x;`. g++ complains about possibly undefined behavior. Replaced with `x += p.x`
- In _random.cpp_ : `if (!left < 0)` is always false according to clang. The correct implementation is probably `if (left < 0)

### DIFF
--- a/src/app/plugins/plugin_dvr.cpp
+++ b/src/app/plugins/plugin_dvr.cpp
@@ -364,7 +364,6 @@ void PluginDVR::slotMovieSave() {
         if (output.getNumBytes() > 0) {
           //write file:
           QString num = QString::number(i);
-          int n = max(5,num.length());
           num = "00000" + num;
           num = num.right(5);
           QString filename = dir + "/" + num + ".png";
--- a/src/app/plugins/plugin_legacypublishgeometry.cpp
+++ b/src/app/plugins/plugin_legacypublishgeometry.cpp
@@ -48,8 +48,8 @@ PluginLegacyPublishGeometry::PluginLegacyPublishGeometry(
     FrameBuffer* fb,
     RoboCupSSLServer* ds_udp_server_old,
     const RoboCupField& field) :
-    _ds_udp_server_old(ds_udp_server_old),
     VisionPlugin(fb),
+    _ds_udp_server_old(ds_udp_server_old),
     _field(field) {
   setSharedAmongStacks(true);
   _settings=new VarList("Publish Legacy Geometry");
--- a/src/graphicalClient/soccerview.cpp
+++ b/src/graphicalClient/soccerview.cpp
@@ -29,6 +29,16 @@ using FieldConstantsRoboCup2014::kNumFieldArcs;
 using FieldConstantsRoboCup2014::kFieldLines;
 using FieldConstantsRoboCup2014::kFieldArcs;
 
+const double GLSoccerView::minZValue = -10;
+const double GLSoccerView::maxZValue = 10;
+const double GLSoccerView::FieldZ = 1.0;
+const double GLSoccerView::RobotZ = 2.0;
+const double GLSoccerView::BallZ = 3.0;
+const int GLSoccerView::PreferedWidth = 1024;
+const int GLSoccerView::PreferedHeight = 768;
+const double GLSoccerView::MinRedrawInterval = 0.016; ///Minimum time between graphics updates (limits the fps)
+const int GLSoccerView::unknownRobotID = -1;
+
 GLSoccerView::FieldDimensions::FieldDimensions() :
   field_length(FieldConstantsRoboCup2014::kFieldLength),
   field_width(FieldConstantsRoboCup2014::kFieldWidth),
--- a/src/graphicalClient/soccerview.cpp
+++ b/src/graphicalClient/soccerview.cpp
@@ -82,8 +82,6 @@ void GLSoccerView::mousePressEvent(QMouseEvent* event)
   leftButton = event->buttons().testFlag(Qt::LeftButton);
   midButton = event->buttons().testFlag(Qt::MidButton);
   rightButton = event->buttons().testFlag(Qt::RightButton);
-  bool shiftKey = event->modifiers().testFlag(Qt::ShiftModifier);
-  bool ctrlKey = event->modifiers().testFlag(Qt::ControlModifier);
 
   if(leftButton)
     setCursor(Qt::ClosedHandCursor);
@@ -99,8 +97,6 @@ void GLSoccerView::mousePressEvent(QMouseEvent* event)
 
 void GLSoccerView::mouseReleaseEvent(QMouseEvent* event)
 {
-  bool shiftKey = event->modifiers().testFlag(Qt::ShiftModifier);
-  bool ctrlKey = event->modifiers().testFlag(Qt::ControlModifier);
   setCursor(Qt::ArrowCursor);
 }
 
@@ -124,7 +120,6 @@ void GLSoccerView::mouseMoveEvent(QMouseEvent* event)
   }else if(midButton){
     //Zoom
     double zoomRatio = double(event->y() - mouseStartY)/500.0;
-    double oldScale = viewScale;
     viewScale = viewScale*(1.0+zoomRatio);
     recomputeProjection();
     mouseStartX = event->x();
@@ -137,7 +132,6 @@ void GLSoccerView::wheelEvent(QWheelEvent* event)
 {
   static const bool debug = false;
   double zoomRatio = -double(event->delta())/1000.0;
-  double oldScale = viewScale;
   viewScale = viewScale*(1.0+zoomRatio);
   recomputeProjection();
   if(debug) printf("Zoom: %5.3f\n",viewScale);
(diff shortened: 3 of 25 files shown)


ezavesky:
These look sensible.  Changes like [this one](https://github.com/RoboCup-SSL/ssl-vision/pull/53/files#diff-0af35ebc8eb7ce9a79873b5b1432a20aR419) will help developers more easily build the app on other platforms, too.

MichaelEischer:
I've rebased the original commit and added a second one that makes the codebase compatible with C++11 while staying compatible with C++98.
@joydeep-b: Any chance to merge this any time soon?

joydeep-b:
@MichaelEischer have you tested this with a four-cam firewire setup to verify that the core functionality is correct?
I have a few questions regarding specific line changes, shall comment inline.

src/app/plugins/plugin_dvr.cpp
```diff
@@ -364,7 +364,6 @@ void PluginDVR::slotMovieSave() {
         if (output.getNumBytes() > 0) {
           //write file:
           QString num = QString::number(i);
-          int n = max(5,num.length());
```
joydeep-b: Could you please explain why this is no longer required?

src/app/plugins/plugin_legacypublishgeometry.cpp
```diff
@@ -48,8 +48,8 @@ PluginLegacyPublishGeometry::PluginLegacyPublishGeometry(
     FrameBuffer* fb,
     RoboCupSSLServer* ds_udp_server_old,
     const RoboCupField& field) :
-    _ds_udp_server_old(ds_udp_server_old),
```
joydeep-b: Could you please explain why this line had to be moved?

src/graphicalClient/soccerview.cpp
```diff
@@ -99,8 +107,7 @@ void GLSoccerView::mousePressEvent(QMouseEvent* event)
 
 void GLSoccerView::mouseReleaseEvent(QMouseEvent* event)
 {
-  bool shiftKey = event->modifiers().testFlag(Qt::ShiftModifier);
-  bool ctrlKey = event->modifiers().testFlag(Qt::ControlModifier);
+  (void)event;
```
joydeep-b: Does removing this result in an unused variable warning? If so, what compiler and version?

src/shared/capture/capture_generator.cpp
```diff
@@ -138,7 +138,6 @@ RawImage CaptureGenerator::getFrame()
     int slice_width = w/n_colors;
     rgb color=RGB::Black;
     rgb color2;
-    int gradient=h/256;
```
joydeep-b: Is this an unused var?

src/shared/capture/capturev4l.cpp
```diff
@@ -818,18 +817,12 @@ void CaptureV4L::readParameterValues(VarList * item) {
...
-    VarBool * vdefault=0;
     VarBool * vwasread=0;
     
-    bool has_abs=false;
-    
     vector<VarType *> children=item->getChildren();
     for (unsigned int i=0;i<children.size();i++) {
         if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="was_read") vwasread=(VarBool *)children[i];
         if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="enabled") venabled=(VarBool *)children[i];
-        if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="default") vdefault=(VarBool *)children[i];
```
joydeep-b: Are these lines removed since the vars are set but not used?

src/shared/capture/capturev4l.cpp
```diff
@@ -884,18 +877,15 @@ void CaptureV4L::writeParameterValues(VarList * item) {
...
     VarBool * vwasread=0;
     VarBool * vdefault=0;
     
     vector<VarType *> children=item->getChildren();
     for (unsigned int i=0;i<children.size();i++) {
         if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="was_read") vwasread=(VarBool *)children[i];
         if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="enabled") venabled=(VarBool *)children[i];
-        if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="auto") vauto=(VarBool *)children[i];
-        if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="default") vauto=(VarBool *)children[i];
+        if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="default") vdefault=(VarBool *)children[i];
```
joydeep-b: Why is "default" written (line 888), bu t not read (line 831)?

src/shared/capture/capturev4l.cpp
```diff
@@ -939,23 +929,13 @@ void CaptureV4L::readParameterProperty(VarList * item) {
...
     VarBool * venabled=0;
-    VarBool * vauto=0;
-    VarInt * vint2=0;
-    VarInt * vint3=0;
-    VarTrigger * vtrigger=0;
     
     vector<VarType *> children=item->getChildren();
     for (unsigned int i=0;i<children.size();i++) {
         if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="enabled") venabled=(VarBool *)children[i];
-        if (children[i]->getType()==VARTYPE_ID_BOOL && children[i]->getName()=="auto") vauto=(VarBool *)children[i];
```
joydeep-b: Could you please explain why these lines have been removed?

src/shared/cmpattern/cmpattern_teamdetector.cpp
```diff
@@ -416,7 +416,7 @@ void TeamDetector::findRobotsByModel(::google::protobuf::RepeatedPtrField< ::SSL
   (void)image;
   const int MaxDetections = _other_markers_max_detections;
   Marker cen; // center marker
-  Marker markers[MaxDetections];
```
joydeep-b: Why did this need to be converted from a stack allocation to a heap allocation?

src/shared/cmpattern/cmpattern_teamdetector.h
```diff
@@ -40,14 +40,6 @@
 using namespace std;
 namespace CMPattern {
 
-class TeamDetectorInterface {
```
joydeep-b: Why did this have to be removed?

src/shared/capture/capturev4l.cpp
MichaelEischer: vdefault is somehow used to reset vint. Not sure what's exactly happening there.

src/shared/capture/capturev4l.cpp
MichaelEischer: Those variables are once again never read. I've just noticed that these changes should be accompanied by a cleanup of the var type tree.

src/shared/cmpattern/cmpattern_teamdetector.h
MichaelEischer: The update function was out of sync and it's not used anywhere else.

src/shared/cmpattern/cmpattern_teamdetector.cpp
```diff
@@ -506,6 +506,7 @@ void TeamDetector::findRobotsByModel(::google::protobuf::RepeatedPtrField< ::SSL
     robots->RemoveLast();
   }
 
+  delete markers;
```
MichaelEischer: This should be `delete[]` . I'm a bit surprised that I didn't notice this given that clang is complaining now.

src/shared/cmpattern/cmpattern_teamdetector.cpp
MichaelEischer: clang complains otherwise:
> variable length array of non-POD element type 'CMPattern::Marker'

POD is short for Plain Old Data. The Marker class has a constructor and thus doesn't count as POD.

src/shared/util/gvector.h
```diff
@@ -212,9 +212,9 @@ vector3d<num> cross(const vector3d<num> a,const vector3d<num> b)
   template <class num> \
   vector3d<num> &vector3d<num>::operator opr (const vector3d<num> p) \
   {                  \
-    x = x opr p.x;   \
```
MichaelEischer: I think it was g++ that complained about these assignments. Took me a while to understand what's happening there.

MichaelEischer:
I've successfully tested the changes with a robot shooting goals in a two camera setup. Currently only a few further cleanups of the capturev4l code remain.

joydeep-b:
Michael, thanks for this changeset.
Looking through the source, it seems that the v4l drivers are sorely in need of cleanup, along with some paring of the associated VarTree. I'll add that as an issue. 
For now, this changeset looks good to me.

PR #54, opened 2016-06-03
Set required packages

Tosainu:
CMake has `find_package()` command to find external libraries.
I changed to use it to check required packages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,14 +8,27 @@ add_definitions(-D _GNU_SOURCE)
 
 SET(EXECUTABLE_OUTPUT_PATH ${PROJECT_SOURCE_DIR}/bin)
 
-set (QT_USE_QTNETWORK true)
-set (QT_USE_QTOPENGL true)
+# Find third-party libraries
+find_package(OpenGL REQUIRED)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -140,13 +153,18 @@ message (STATUS "Custom CC Flags: ${cc_flags}")
 
 add_definitions(${cc_flags})
 
-set (libs ${QT_LIBRARIES} ${EIGEN3_LIBRARIES} dc1394 jpeg png protobuf pthread GL GLU )
-
 ## build the common code
 add_library(sslvision ${SHARED_MOC_SRCS} ${SHARED_RC_SRCS} ${CC_PROTO} ${SHARED_SRCS})
-add_dependencies(sslvision GenerateProto )
-target_link_libraries(sslvision ${libs})
```
justbuchanan: Should this line be added back in after `libs` is set?

Tosainu: It's not necessary.
> If a static library's code contains references to some shared library items, these references will become dependencies in the resulting executable.
http://stackoverflow.com/a/7841968

joydeep-b:
Looks good to me.

PR #56, opened 2016-07-26
added high resolution timer from c++11

g3force:
A year or two ago, I applied this change to our vision and what to share it now.

Without this change, the timestamps I received from SSL vision in our AI software diverged from our internally tracked time (Java's System.nanotime() ). With the high resolution clock, time measurement is more precise.

The only drawback of this change is the use of C++11, but this should not be a problem anymore today...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ cmake_minimum_required(VERSION 2.6)
 add_definitions(-D UNIX)
 add_definitions(-D LINUX)
 add_definitions(-D _GNU_SOURCE)
+add_definitions(-std=c++11 -fpermissive)
 
 SET(EXECUTABLE_OUTPUT_PATH ${PROJECT_SOURCE_DIR}/bin)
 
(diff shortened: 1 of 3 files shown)


joydeep-b:
This can be done without the additional requirement of C++11, by using clock_gettime. The move to C++11 should be a separate issue, up for debate, due to the additional dependencies on the compilers, and hence the systems, that it imposes.

CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ cmake_minimum_required(VERSION 2.6)
 add_definitions(-D UNIX)
 add_definitions(-D LINUX)
 add_definitions(-D _GNU_SOURCE)
+add_definitions(-std=c++11 -fpermissive)
```
MichaelEischer: Using -fpermissive is not particularly nice. This commit seems to indicate that supporting C++11 requires some code changes besides changed compiler flags. Thus it would be much nicer too fix all compile errors instead of disabling them.

MichaelEischer: I've updated my pull request #53 to make the ssl-vision codebase compatible with both C++98 and C++11.

g3force:
I remember that I tried clock_gettime first, which did not work out. But this change was done about 2 or 3 years ago... I will verify if time still drifts in the current master later.

g3force:
I can not reproduce the issue anymore, so this change is obsolete.

PR #58, opened 2016-08-04
Added Neural Network color calibration

g3force:
The neuralcolorcalib plugin was written by 
J. A. Gurzoni Jr and Gabriel Francischini
and was originially published as a downloadable archive on googlecode

I added the plugin to the git repo without modifications.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,8 +26,6 @@ find_package(PkgConfig REQUIRED)
 pkg_check_modules(Eigen3 REQUIRED eigen3)
 pkg_check_modules(DC1394 REQUIRED libdc1394-2)
 
-find_package( OpenCV REQUIRED )
-
 include_directories(
 	${PROJECT_BINARY_DIR}
 	${Eigen3_INCLUDE_DIRS}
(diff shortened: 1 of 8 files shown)


joydeep-b:
This is a valuable commit, but it has a significant dependency: OpenCV. I think it would be better to make this an optional dependency. If opencv is detected, the plugin could be compiled and added, else it should print a message in the cmake setup and disable compiling and adding the plugin.

g3force:
OpenCV is now optional.

jgkamat:
Thanks @g3force this works very nicely! (anyone trying to set this up, install `libopencv-dev` in ubuntu). If possible, it would be nice to add these features (but these could be done at a later date):

1. Undo addition of training points (at least one)
2. Save/Restore of Training Data
3. Save state of LUT so we can undo from a neural network color calibration (it's a little startling to have your config overwritten)

I'd be happy to write a wiki article on getting this set up (which I'll work on probably next week).

Thanks so much, this is a huge improvement over the old color calibration method!

joydeep-b:
Looks good! Merging into master now.

g3force:
jgkamat, you should thank the developers, not me ;) but I found it really annoying that this feature was never merged into master or any other branch... Nobody knew about this work, although it is already some years old.
By the way, I am also working on some online calibration, but this is still work in progress...

joydeep-b:
This PR breaks the build on systems which have OpenCV versions >= 3 installed.
#63

PR #60, opened 2016-10-15
Fixes some memory leaks in the software

hbritto:


src/app/stacks/multivisionstack.cpp
```diff
@@ -27,6 +27,9 @@ MultiVisionStack::MultiVisionStack(string _name, RenderOptions * _opts) {
 }
 
 MultiVisionStack::~MultiVisionStack() {
+  for (int i=0;i<threads.size();i++) {
```
MichaelEischer: See comment on mainwindow.cpp

src/graphicalClient/soccerview.cpp
```diff
@@ -170,6 +170,8 @@ void GLSoccerView::resizeEvent(QResizeEvent* event)
 {
   QGLWidget::resizeEvent(event);
   redraw();
+
+  resetView();
```
MichaelEischer: What is that supposed to do? As far as I can tell this has nothing to do with fixing memory leaks but changes the behavior when resizing the window. If i remember correctly, the old behavior was that resizing the window doesn't change the size of the rendered field. The change would always reset the field size to its default when the window is resized.
Why still redraw the widget before resetView is called? The latter calls postRedraw() which queues a call to redraw in the Qt event loop.

src/app/gui/mainwindow.cpp
```diff
@@ -227,5 +227,6 @@ MainWindow::~MainWindow() {
 
   // Stop stack:
   multi_stack->stop();
+  delete multi_stack;
```
hbritto: As my camera uses the Ethernet port to communicate, me and my team made a plugin based on the existing ones for it. The thing is, we use the Vimba API, and every time ssl-vision is closed, it gives a segmentation fault. By adding this, and the deletes in `multivisionstack.cpp`, we found that the error was fixed.

CMakeLists.txt
```diff
@@ -123,10 +123,10 @@ message (STATUS "Arch: ${arch}")
 
 if (${arch} MATCHES "x86_64")
 	#set compiler flags here for x86_64
-	set (cc_flags "")
+        set (cc_flags "-ansi")
```
MichaelEischer: Sorry about complaining over that flag, should have tried compiling with C++11 myself before.
So here's what's happening: GCC 6 changed the C++ default from gnu++98 to gnu++14. The -ansi flag reverts this change and explicitly states the intended C++ version.
I would prefer that the flags are set using the following code snippet just after the if containing the set cc_flags commands.
<pre>
# make sure to use C++98
SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ansi")
</pre>

Another option would be to use the commits from my pull request #53 which adds support for both C++ versions.

src/app/stacks/multivisionstack.cpp
joydeep-b: It would be safest to call stop() and then wait() on each thread before deleting them.

joydeep-b:
Once the change to stop() and wait() are made in the destructor, this should be ready to merge.

src/app/stacks/multivisionstack.cpp
```diff
@@ -27,6 +27,11 @@ MultiVisionStack::MultiVisionStack(string _name, RenderOptions * _opts) {
...
+	threads.at(i)->stop();
```
joydeep-b: Sorry, I meant MultiVisionStack::stop(), which should take care of shutting down all the threads.

joydeep-b:
Thanks for the pull-request, it is now merged!

PR #67, opened 2016-11-24
set tSent within mutex when sending cam frames

g3force:
* this ensures that tSent will be strictly monotonically increasing among all cameras

### DIFF
--- a/src/shared/net/robocup_ssl_server.cpp
+++ b/src/shared/net/robocup_ssl_server.cpp
@@ -19,6 +19,7 @@
 */
 //========================================================================
 #include "robocup_ssl_server.h"
+#include "timer.h"
 
 RoboCupSSLServer::RoboCupSSLServer(int port,
                      string net_address,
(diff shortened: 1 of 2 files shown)


MichaelEischer:
Is there any guarantee on the delivery order of the camera frames?

g3force:
No, there is no guarantee.
For testing purposes, I used a global frame id that I set after tSent (inside mutex lock). It was always in-order during the measurement period of some seconds.
The client was on the same machine, though...

We receive all camera frames into one list and calculate the time difference between two consecutive frames and it turned out that we get negative dTs for both tCapture and tSent.
While it is reasonable for tCapture (unpredictable processing time, threading) it is strange for tSent.

joydeep-b:
Could you please explain what problem this solves? It would especially help if you could share some minimal code to re-create the problem as a standalone executable. Could you modify the client to do this? Perhaps remove all the processing and printing, except for the essential parts to demonstrate the problem?

Note that UDP packets can arrive out of order anyway, so it is not possible to ensure that the order that they are sent will be the same as the order received.

Deferring serial operations until absolutely necessary allows computational speedup from parallel execution. Adding a mutex to these operations makes them serial, where the packing / marshalling each separate packet could have been done in parallel.

g3force:
There is not a major problem, I just stumbled upon this issue. I find it wrong that tSent is not the timestamp just before sending.
I did some more detailed analysis. During normal operation, there are actually only a few frames that are out of order. With the patch, there are none. However, as soon as there is some more load on the system (I did a stress -c 2 which puts 100% load on 2 CPU cores), you can see that the calculated DT of tSent is negative quite often. That is reasonable, as the system has a quad core without hyperthreading and the SSL vision uses 4 cameras.
I could not measure any major performance drops by the way.
Attached you will find a patch for the client that captures a log file.
I also added some figures, some log files and a matlab script.
From my point of view it is safe to apply the changes, but I would be fine without it, too. We just need to make sure to always have more or equal cores than cameras.

[vision-stuff.zip](https://github.com/RoboCup-SSL/ssl-vision/files/637523/vision-stuff.zip)

joydeep-b:
@g3force  Thanks for sharing the detailed information! I think I get it now. Please correct my summary if it is wrong:

When packets are received by a client on the same computer, there are occasionally successive packets (say, p1, p2, where p2 is the packet received just after p1) such that (p2.tSent - p1.tSent) < 0. 

From your graphs, it seems like this occurs at a rate ~ 1 in 10^4 packets when there are at least as many cores available to ssl-vision as there are cameras. When there are fewer cores available than cameras, the occurrence is much higher.

Okay, looks good, shall test on a different system, and if successful should merge.

PR #68, opened 2016-12-09
InstallPackage<OS>.sh scripts

kylevedder:
I have converted the lines in the README listing what packages to install into scripts that can be easily run.

This makes the codebase easier to get up and running without the need to copy and paste specific lines of the README into the console.

### DIFF
--- a/InstallPackagesArch.sh
+++ b/InstallPackagesArch.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+# These are the packages required for install on Arch Linux systems.
+sudo pacman -Sy gcc qt4 eigen protobuf libdc1394 cmake v4l-utils jsoncpp
--- a/InstallPackagesUbuntu.sh
+++ b/InstallPackagesUbuntu.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+# These are the packages required for install on Ubuntu systems.
+sudo apt-get install g++ libqt4-dev libeigen3-dev protobuf-compiler libprotobuf-dev libdc1394-22 libdc1394-22-dev cmake libv4l-0
--- a/README.md
+++ b/README.md
@@ -40,14 +40,9 @@
  * libpng
  * video for linux 2 (v4l)
 
-To get all of these packages in (k)ubuntu, run:
-```
-    sudo apt-get install g++ libqt4-dev libeigen3-dev protobuf-compiler libprotobuf-dev libdc1394-22 libdc1394-22-dev cmake libv4l-0
-```
-Or in archlinux, run:
-```
-$ pacman -Sy gcc qt4 eigen protobuf libdc1394 cmake v4l-utils jsoncpp
-```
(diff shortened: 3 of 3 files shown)


MichaelEischer:
I don't see the point in moving these lines into separate files. You still want to look at the scripts before running them. And after all copying one command line isn't too hard. In addition the user must already have looked at the readme in order to know how to complete the build process.

kylevedder:
I see your point about not knowing what's in the script; however, the user can `cat` it.

I don't think you need to look at the README to be able to build the tools. Simply seeing the Makefile was enough for me to know to run `make`; the only stumbling block was not having all of the dependencies installed on my machine, which would be easy to remedy if I also see `InstallPackages<OS>.sh` as well.

In addition, this allows seasoned users to get the SSL Vision system up and running more easily as they don't need to open the README file and copy the line into their terminal before use, whereas they would with the current README. @joydeep-b currently uses this style of install packages script in his RoboCup SSL repo and I find it very convenient to use.

Finally, I also submitted this because Joydeep said he thought it was a good idea and that he would accept my PR.

MichaelEischer:
You're right that seasoned users wouldn't need to open the readme. The only exception would be the usage of the maxtrixvision cameras, but I think that would be fine.

kylevedder:
I am working on squashing my commits.

kylevedder:
OK, commits squashed. Thank you for the suggestion.

InstallPackagesArch.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# These are the packages required for install on Arch Linux systems.
+# Taken from the README.
+pacman -Sy gcc qt4 eigen protobuf libdc1394 cmake v4l-utils jsoncpp
```
justbuchanan: This probably needs a `sudo` at the beginning.

kylevedder: OK, noted. Just as an FYI, this is how it currently stands in the README. I will fix that now.

kylevedder: Fixed as requested.

joydeep-b:
@MichaelEischer : This is just a convenience script, to streamline setup. Side-note, If you don't trust the contents of a script in a repo, you shouldn't trust the code from the repo, either!  :)

@kylevedder : Adding some additional change requests.

InstallPackagesArch.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# These are the packages required for install on Arch Linux systems.
+# Taken from the README.
```
joydeep-b: "Taken from README" can be removed.

InstallPackagesUbuntu.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# These are the packages required for install on Ubuntu systems.
+# Taken from the README.
```
joydeep-b: "Taken from README" can be removed.

kylevedder: Noted. I will change and push now.

Issue #76, opened 2017-07-22
Problem detecting patterns


maxaubel:
Hi!
Because of the room size, we made a field 5 times smaller, we also had to scale down the robot sizes. Long story short, we detect robots, but we can't recognize the single patterns from a team. 

When "Load image files" is not selected, this is what I see:
![image](https://user-images.githubusercontent.com/28493557/28494865-43fe3540-6f09-11e7-88e3-17f9cd843f10.png)
It gives exactly the position of every robot and its team color, but we cannot tell which robot specifically it is. In that matter, we needed to load the pattern in order to identify the robots.

This is how we have configured the blue team (when telling it to tell load image file):
![image](https://user-images.githubusercontent.com/28493557/28494961-4e30f086-6f0c-11e7-8d0c-93157ddb5cf1.png)
As You can see, in terminal it printed that it couldn't load the image file, I don't know why.

And when it's configured as said previously, this is what I see in graphicalClient:
![image](https://user-images.githubusercontent.com/28493557/28494868-6aade08c-6f09-11e7-8307-6d227a8bff15.png)

In the pattern directory, there's the file blue.png and blue.svg containing the patterns of the blue team (and there's one for the yellow team too).

Any suggestions on that? I have set that the camera height is 5 meters, because the camera is at about 1 meter from the field. By the way, we currently use a Stingray F046C.

jgkamat:
Hi! I'll start by saying I'm no expert in the vision system, but hopefully this advice will help.

I find that the vision templates don't matter too much, especially if you don't stray too far from the standard design. I would suggest selecting the "standard2010" team, since that's preconfigured and should work right away.

A common mistake (and a frustration of mine) is that SSL vision uses paths relative to the directory you run SSLVision from. This means that if you run SSLVision from a directory it's not expecting, all the paths will be completely off. I noticed you're running SSLVision from the `ssl-vision/bin/` directory, which means that the default configuration for vision won't be loaded! To fix loading default teams, and fix loading your images, you should run SSLVision like this:

```
# This depends on where you have SSLVision cloned, of course
cd ~/Desktop/ssl-vision
./bin/vision
```

NOTE: This will loose all configuration that you previously made, because SSLVision also stores configuration files relative to the current working directory. If you want to keep your config, you can move the xml files generated from `~/Desktop/ssl-vision/bin` to `~/Desktop/ssl-vision`, BUT I highly reccomend you start with a fresh config, so you don't have to do extra debugging.

[...]

PR #78, opened 2017-08-06
Add a warning when running vision from a non-standard directory

jgkamat:
During RoboCup 2017, there were two independent instances where someone ran vision from the `./bin` folder, which lost all configuration, and caused a lot of confusion.

There's also the problem in #76, where the config system has a bunch of variables that expect the root of the repo to be the cwd.

To try to solve both of these problems, I've added a warning which will print when running from a directory other than the root of the repo. It's a little bit ugly, unfortunately, because I don't think ssl-vision uses boost (so I can't use `boost::filesystem`), but it's a start.

Let me know if you have any feedback :smile:

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -26,6 +26,8 @@
 #include "mainwindow.h"
 #include <signal.h>
 #include <stdio.h>
+#include <iostream>
+#include <unistd.h>
 #include "qgetopt.h"
 
 MainWindow* mainWinPtr = NULL;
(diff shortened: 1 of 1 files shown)


src/app/main.cpp
```diff
@@ -39,6 +41,34 @@ void HandleStop(int i) {
...
+  int count = readlink("/proc/self/exe", binaryPathRaw, PATH_MAX);
+
+  if (count == -1) {
+    // We failed getting the binary path, quit
+    return;
+  }
+  std::string binaryPath = std::string(binaryPathRaw, count);
+
+  // Get the offset to check
+  unsigned int offset = binaryPath.size() - (binaryPath.size() - currentWorkingDir.size());
```
JetseVerschuren: Change `unsigned int offset = binaryPath.size() - (binaryPath.size() - currentWorkingDir.size());` to `unsigned int offset = currentWorkingDir.size();`.
It does the same, just simplified.

JetseVerschuren: Change `unsigned int offset = binaryPath.size() - (binaryPath.size() - currentWorkingDir.size());` to `unsigned int offset = currentWorkingDir.size();`
It does the same.

jgkamat: thanks, nice catch, I probably need more sleep...

src/app/main.cpp
```diff
@@ -39,6 +41,34 @@ void HandleStop(int i) {
   }
 }
 
+// Print a path warning if the running directory is NOT one level above the binary directory
+void printPathWarning() {
+  char cwd[PATH_MAX];
+  getcwd(cwd, PATH_MAX);
```
joydeep-b: The return value of getcwd needs to be checked for failure:
https://linux.die.net/man/3/getcwd

jgkamat:
Nice catch on the `getcwd` usage, that definetly needs to be fixed :)

I think that exiting immediately is not the right thing to do, in case someone knows what they are doing. For example, RoboJackets intentionally runs `vision` from the `bin` directory, so that we can make the `./bin` folder it's own submodule and [publish our config files](https://github.com/RoboJackets/ssl-vision-config). 

I personally think that a console warning will be enough: in all the cases where I had this problem (both at competition, and in the issue referenced above), people checked the console when they realized something was wrong. 

I think that a gui dialog is a good compromise though, so I'll add that in for now. Maybe in the future I'll add another PR creating a setting to squelch the warning.

joydeep-b:
Looks good, thanks!

PR #81, opened 2017-09-17
Add section for other clients

ivanseidel:
I believe many teams will get benefited from more tools and clients to SSL Vision or other vision based competitions as well being put as simple as possible to get started quickly.


g3force:
I have added the change to the Wiki: https://github.com/RoboCup-SSL/ssl-vision/wiki/communication

We should keep the Readme short with only the relevant information for running ssl-vision.

PR #82, opened 2017-12-11
Choose a random free port instead of the target port for local bind

g3force:
If some other applications also binds to the same local port and forgets to also set SO_REUSEPORT, the bind will fail.
There is no reason for binding to a fixed local port in this case.


joydeep-b:
This looks fine to me, once tested, can merge to master.

asctec-andre:
I tested this with a real field and vision setup and can confirm that the vision now uses an ephemeral source port. Target port and functionality is unaffected and works as before.

PR #47, opened 2016-02-10
Feature/lut copy

ezavesky:
**NOTE** This is a refactored pull request from #41 that excludes team pattern changes.
-  add button in "YUV Calibrator" to copy LUT from other cameras (solving [legacy request](https://github.com/RoboCup-SSL/ssl-vision/blob/wiki/TODO.md) to allow easy LUT channel sync)
  - **NOTE** builds on some vartype changes in PR #39 
  - a space to limit which color is copied, but not currently implemented (let's gauge the interest first)
- update vartype to allow access to parent node

160207_lut_copy

PR #84, opened 2017-12-19
Lut copy

g3force:
based on the work in #47


g3force:
LUT copy with color selection works now.
Can someone review and test the changes?

joydeep-b:
@sdlane will run tests in our lab.

sdlane:
This works well on our system. Everything functions as I would expect and nothing seems to interfere with normal operations. This seems like it will be incredibly useful for re-calibrating and especially new calibration.

PR #88, opened 2018-01-14
 Simplify camera calibration setup

g3force:
- generate lines and arcs from field parameters
- generate control points from camera index
- support more cameras than attached locally by specifying the global camera id and sending this instead in the detection frame
- generate rectangular penalty area for RoboCup 2018


src/shared/util/field_default_constants.h
```diff
@@ -29,20 +29,16 @@
...
-const double kBoundaryWidth = 250.0;
-
-extern const std::size_t kNumFieldLines;
-extern const FieldLine kFieldLines[];
-
-extern const std::size_t kNumFieldArcs;
-extern const FieldCircularArc kFieldArcs[];
-
-extern const GVector::vector2d<double> kCameraControlPoints[4][4];
+namespace FieldConstantsRoboCup2018A {
```
mahi97: is this division A parameters ?
If it's division-A, field length, field width, and goal width should be change to new numbers.

src/shared/util/camera_calibration.cpp
```diff
@@ -775,6 +759,70 @@ CameraParameters::AdditionalCalibrationInformation::
...
+    }
+
+    for (int i = 0; i < kNumControlPoints; ++i) {
+        control_point_field_xs[i]->setDouble(control_points[i].x);
+        control_point_field_ys[i]->setDouble(control_points[i].y);
+    }
+}
+
+std::vector<GVector::vector2d<double> > CameraParameters::AdditionalCalibrationInformation::
+generateCameraControlPoints(int cameraId, int numCamerasTotal, double fieldHeight, double fieldWidth) {
```
asctec-andre: This is a very helpful thing. It was never easy to get the control points right, even for experienced users.
Could you add images to the wiki which show the field and 1-8 cameras with their ID? I assume this helpful function will not work properly if one does not know which camera ID should be where.

src/shared/util/field_default_constants.h
g3force: oh yes, sry. I missed that.
I've updated it with the correct dimensions for div A. The dimension of the center circle is not clear yet, but at least I moved it to the constants too.
Anyway, this are only the defaults when initially starting the vision.

src/shared/util/camera_calibration.cpp
g3force: I added a picture to the wiki: https://github.com/RoboCup-SSL/ssl-vision/wiki/camera-calibration

I will also update the rest of the Wiki page later.

asctec-andre: Awesome, thanks!

g3force: Documented the rest as well now: https://github.com/RoboCup-SSL/ssl-vision/wiki/camera-calibration

g3force:
Found a bug in the control point calculation and fixed it.

PR #106, opened 2018-08-28
Add Basler support to SSL-Vision

emielsteerneman:
With this pull request I'd like to add Basler support to SSL-vision. It requires Basler software _Pylon_ to be installed, which can be found [here](https://www.baslerweb.com/en/sales-support/downloads/software-downloads/).


emielsteerneman:
The folder `build` doesn't exist. adding `mkdir $(buildDir)` to the `Makefile` could fix this.

PR #108, opened 2018-09-07
Fix build issues on mac os

lordhippo:


g3force:
looks good to me and still works on my computer with Bluefox camera.

Could you also update the requirements in the README and point out the limitations of using OSX or at least list what's tested?

PR #117, opened 2019-01-13
[#116] Split camera images and distribute them as a virtual source

g3force:


g3force:
works so far for arbitrary cameras, even for the fromFile source.

Next, I need to find or generate some test data to test geometry calibration and stuff.

src/app/stacks/multistack_robocup_ssl.cpp
```diff
@@ -84,9 +86,11 @@ MultiStackRoboCupSSL::MultiStackRoboCupSSL(RenderOptions * _opts, int cameras) :
       *global_field);
 
   //add parameter for number of cameras
-  createThreads(cameras);
-  unsigned int n = threads.size();
-  for (unsigned int i = 0; i < n;i++) {
+  int numNormalCameraThreads = 4;
```
joydeep-b: This is fine for now for testing, but we really should move this to a config param rather than a hard-coded constant.

g3force: where would you put such a parameter? Based on this number, the camera tabs and config tree are populated. Doing this during runtime would probably be a bit more effort.

src/app/capture_thread.cpp
```diff
@@ -260,12 +267,13 @@ void CaptureThread::run() {
                 stack->updateTimingStatistics();
                 stack_mutex.unlock();
               }
-              capture_mutex.lock();
-              if ((capture != 0) && (capture->isCapturing())) {
-                capture->releaseFrame();
-              }
-              capture_mutex.unlock();
           }
+
```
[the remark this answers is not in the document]
g3force: yes. Before, the `releaseFrame` was not called, when `copyAndConvertFrame` returned false. But the `getFrame` may still have allocated the frame already. The new splitter gets out of sync, when there is not a call to `releaseFrame` for every `getFrame`.
It's probably good to test this with a Firewire camera, just to be sure.

g3force:
I was also especially uncertain about the geometry calibration. That's why I did some crazy setup:
img_20190113_205423

This is a USB2 bluefox directed to a 27" monitor showing a 12x9m field. I configured the splitter to a single quadrant and did the geometry calibration:
2019-01-13-205350_773x606_scrot

Looks good :) The camera center seems reasonable. That's the best I can do for now.

joydeep-b:
That looks good! The red cross shows the principal point is correctly found to be in the corner of the image, as expected.
Extra points for the crazy test setup. :+1:

src/app/stacks/multistack_robocup_ssl.cpp
joydeep-b: Config file or command-line params seem like reasonable compromises. Adding it to the GUI parameter tree will require some serious refactoring since this is required during initialization.
Again, it's fine for now.

src/shared/capture/capture_splitter.cpp
```diff
@@ -0,0 +1,188 @@
...
+  // calculate offset in src image
+  int height_offset = (int) (src.getHeight() * rel_height_offset);
+  int width_offset = (int) (src.getWidth() * rel_width_offset);
+
+  // calculate dimensions of target image
+  int height = (int) (src.getHeight() * rel_height);
+  int width = (int) (src.getWidth() * rel_width);
+
+  // for some reason, the image gets screwed when the width is not a multiple of 4
+  width -= width % 4;
```
g3force: @joydeep-b Any idea, why images get screwed with certain width? Looks like it's mostly fine with this hack, but I already saw another glitch with this hack (colors have changed).
We already experienced this when changing the AOI of the bluefox cameras (which is directly set in the cameras). So I do not expect this is an issue during conversion, but rather during visualization.
The visualizations, like field calibration result, is screwed as well.

Example:
2019-01-14-215712_414x369_scrot

joydeep-b: OpenGL Texture operations require specific alignment. The easiest way to do this is to ensure that the image dimensions are in multiples of 4. Or else, you'll need to manually create a new image texture with padding to the next multiple of 4. See:
https://www.khronos.org/opengl/wiki/Common_Mistakes#Texture_upload_and_pixel_reads

g3force: oh, I see. Thx for the hint :)

src/app/stacks/multistack_robocup_ssl.cpp
g3force: I've created a new PR: #127 

Regarding this parameter: I've added a cli parameter to set the number of camera threads. There is one issue though: The VarTree does not delete any nodes automatically. So there will be cameras left, if running with less camera threads then before. There will still be only the given number of threads and tabs, but more elements in the config tree...

PR #118, opened 2019-01-20
WIP: High res camera support

g3force:
Branch, containing overall changes to get high resolution cameras running.
This PR is not indented to be directly merged into master. It's just for reference. I'll split this into multiple PRs later.


g3force:
Debayering a full image (2448x2048) with OpenCV takes 1.9ms on average with this code:
```
cv::Mat srcMat(height, width, CV_8UC1, data_buf);
cv::Mat dstMat(height, width, CV_8UC3, target.getData());
cvtColor(srcMat, dstMat, cv::COLOR_BayerBG2RGB);
```
Doing this on a quarter of the image takes 1.2ms.
Both on a Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz

joydeep-b:
Can you profile that under valgrind to see if there is any large-scale memory allocation / deallocation happening? We should be able to reuse the same source / dest image buffers for each frame, to eliminate memory allocation slowdowns.

g3force:
the opencv conversion does not allocate any extra memory, according to valgrind.
I'm not sure how to check the rest of the vision code. Running valgrind on the vision executable produces a lot of output, then it crashes. I haven't expected much more, though, anyways... =D

PR #125, opened 2019-02-24
Read raw images from files

### LINKED ISSUE #123: Read raw images from files

g3force:
Resolves #123

### DIFF
--- a/src/shared/capture/capturefromfile.cpp
+++ b/src/shared/capture/capturefromfile.cpp
@@ -46,6 +46,7 @@ CaptureFromFile::CaptureFromFile(VarList * _settings) : CaptureInterface(_settin
   conversion_settings->addChild(v_colorout=new VarStringEnum("convert to mode",Colors::colorFormatToString(COLOR_YUV422_UYVY)));
   v_colorout->addItem(Colors::colorFormatToString(COLOR_RGB8));
   v_colorout->addItem(Colors::colorFormatToString(COLOR_YUV422_UYVY));
+  v_colorout->addItem(Colors::colorFormatToString(COLOR_RAW8));
 
   conversion_settings-> addChild(v_raw_width=new VarInt("raw width", 0));
   conversion_settings-> addChild(v_raw_height=new VarInt("raw height", 0));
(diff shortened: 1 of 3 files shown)


src/shared/capture/capturefromfile.cpp
```diff
@@ -119,18 +126,40 @@ bool CaptureFromFile::startCapture()
...
-    {
-      int width(-1);
-      int height(-1);
-      rgba* rgba_img = ImageIO::readRGBA(width, height, currentImage->c_str());
-      fprintf (stderr, "Loaded %s \n", currentImage->c_str());
-      images.push_back(rgba_img);
-      heights.push_back(height);
-      widths.push_back(width);
-      ++currentImage;
+    for (auto &currentImage : imgs_to_load) {
```
joydeep-b: This can be a `const auto&` for specificity, and to help catch accidental modification bugs in the future.

src/shared/capture/capturefromfile.cpp
```diff
@@ -119,18 +126,40 @@ bool CaptureFromFile::startCapture()
...
+        raw_img.allocate(ColorFormat::COLOR_RAW8, width, height);
+        memcpy(raw_img.getData(), buffer.data(), static_cast<size_t>(raw_img.getNumBytes()));
+
+        images.push_back(raw_img);
+      }
+      else
+      {
+        cv::Mat srcImg = imread(currentImage, cv::IMREAD_COLOR);
+        RawImage raw_img;
+        raw_img.allocate(ColorFormat::COLOR_RGB8, width, height);
```
joydeep-b: This seems to assume a specific raw format. If so, `RAW` should be renamed to, for example, `RAWBGR8`.

src/shared/capture/capturefromfile.cpp
```diff
@@ -142,19 +171,31 @@ bool CaptureFromFile::startCapture()
   return true;
 }
 
-bool CaptureFromFile::isImageFileName(const std::string& fileName)
+std::string CaptureFromFile::getFileEnding(const std::string& fileName)
```
joydeep-b: I think you mean `getFileExtension`

src/shared/capture/capturefromfile.cpp
g3force: this is actually the 'else' branch. imread from OpenCV is used to read any non-raw image.
I encountered a bug here anyway and fixed it. The conversion was wrong. OpenCV uses BGR, ssl-vision uses RGB format.
PNG files are now read correctly again.

For the raw files, '.raw' is usually the default extension. spinview generates files with this extension, so I don't want to change that.
I could add another parameter for setting the channel size (8,12,16,32), but everything except 8 is rather unusual...

PR #126, opened 2019-02-24
Add Spinnaker support

### LINKED ISSUE #122: Support for FLIR Blackfly via Spinnaker SDK

g3force:
Resolves #122 

There is still an open issue that I reported to ptgrey - except from this, ready to be reviewed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,10 @@
 project(ssl-vision)
 set(OpenGL_GL_PREFERENCE LEGACY)
-set(USE_NEURAL_CALIBRATION FALSE CACHE BOOL "Enable neural network calibration (requires OpenCV 2, not compatible to later versions)")
-set(USE_DC1394 TRUE CACHE BOOL "Compile with DC1394 (firewire cameras)")
 set(USE_QT5 FALSE CACHE BOOL "Use Qt5 instead of Qt4")
-set(USE_mvIMPACT FALSE CACHE BOOL "Compile with mvImpact (bluefox cameras)")
+set(USE_NEURAL_CALIBRATION FALSE CACHE BOOL "Enable neural network calibration (requires OpenCV 2, not compatible to later versions)")
+set(USE_DC1394 TRUE CACHE BOOL "Compile with DC1394 driver (firewire cameras)")
+set(USE_SPINNAKER FALSE CACHE BOOL "Compile with Spinnaker driver (FLIR USB3 cameras)")
+set(USE_mvIMPACT FALSE CACHE BOOL "Compile with mvImpact driver (bluefox USB2 + USB3 cameras)")
 
 if(USE_DC1394 AND USE_mvIMPACT)
 	message(FATAL_ERROR "DC1394 and mvImpact are not compatible: mvImpact crashes when creating device manager")
@@ -175,6 +176,18 @@ if(OpenCV_FOUND)
   endif()
 endif()
 
+if( USE_SPINNAKER )
+	find_package( SPINNAKER REQUIRED)
+	add_definitions( -DSPINNAKER )
+	include_directories( ${SPINNAKER_INCLUDE_DIRS} )
+	set (SHARED_SRCS ${SHARED_SRCS} ${shared_dir}/capture/capture_spinnaker.cpp)
+	set (SHARED_HEADERS ${SHARED_HEADERS} ${shared_dir}/capture/capture_spinnaker.h)
+	message(STATUS "SPINNAKER support enabled")
+else()
+	set( SPINNAKER_LIBS "" )
+	message(STATUS "SPINNAKER support disabled")
+endif()
+
 include_directories(${PROJECT_SOURCE_DIR}/src/app)
 include_directories(${PROJECT_SOURCE_DIR}/src/app/gui)
 include_directories(${PROJECT_SOURCE_DIR}/src/app/plugins)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -1,9 +1,10 @@
 project(ssl-vision)
 set(OpenGL_GL_PREFERENCE LEGACY)
-set(USE_NEURAL_CALIBRATION FALSE CACHE BOOL "Enable neural network calibration (requires OpenCV 2, not compatible to later versions)")
-set(USE_DC1394 TRUE CACHE BOOL "Compile with DC1394 (firewire cameras)")
 set(USE_QT5 FALSE CACHE BOOL "Use Qt5 instead of Qt4")
-set(USE_mvIMPACT FALSE CACHE BOOL "Compile with mvImpact (bluefox cameras)")
+set(USE_NEURAL_CALIBRATION FALSE CACHE BOOL "Enable neural network calibration (requires OpenCV 2, not compatible to later versions)")
```
joydeep-b: Should be "compatible *with*"

src/app/capture_thread.cpp
```diff
@@ -87,6 +87,12 @@ CaptureThread::CaptureThread(int cam_id)
   captureBlueFox3 = new CaptureBlueFox3(bluefox3,camId);
 #endif
 
+#ifdef SPINNAKER
+  captureModule->addItem("Spinnaker");
+  settings->addChild( (VarType*) (spinnaker = new VarList("Spinnaker")));
```
joydeep-b: Recommend splitting up into two lines for readability.
```cpp
spinnaker = new VarList("Spinnaker");
settings->addChild(reinterpret_cast<VarType*>(spinnaker));
```

src/shared/capture/capture_spinnaker.cpp
```diff
@@ -0,0 +1,398 @@
...
+#endif
+{
+  cam_id = (unsigned int) default_camera_id;
+  is_capturing = false;
+
+  #ifndef VDATA_NO_QT
+    mutex.lock();
+  #endif
+
+  settings->addChild(capture_settings = new VarList("Capture Settings"));
```
joydeep-b: Please split into two statements for clarity.

src/shared/capture/capture_spinnaker.cpp
```diff
@@ -0,0 +1,398 @@
...
+  settings->addChild(dcam_parameters  = new VarList("Camera Parameters"));
```
joydeep-b: Ditto.

src/shared/capture/capture_spinnaker.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+  //=======================CAPTURE SETTINGS==========================
+  capture_settings->addChild(v_cam_bus          = new VarInt("cam idx",default_camera_id));
```
joydeep-b: Ditto.

src/shared/capture/capture_spinnaker.cpp
```diff
@@ -0,0 +1,398 @@
...
+  capture_settings->addChild(v_colorout         = new VarStringEnum("color mode", Colors::colorFormatToString(COLOR_RAW8)));
```
joydeep-b: Ditto.

src/shared/capture/capture_spinnaker.cpp
g3force: I agree that it looks nicer when splitted. On the other hand, all capture classes do it like this. I wanted to be consistent with the rest of the vision code.

g3force:
> Looks fine, some small changes requested.
> What is the Spinnaker bug that we are waiting on?

When using the StreamBufferHandlingMode "First Only", reading images from the camera blocks after a short time for ever. spinview (the Test Software of the SDK) does the same.
The "First Only" Mode would be the ideal one for us.
I'm currently in contact with the support of ptgrey.

src/shared/capture/capture_spinnaker.cpp
joydeep-b: Okay, let's go for consistency.

joydeep-b:
> The "First Only" Mode would be the ideal one for us.
> I'm currently in contact with the support of ptgrey.

Is it possible to reduce the buffer size? Then we should be able to reduce the effect of the delay.

g3force:
> > The "First Only" Mode would be the ideal one for us.
> > I'm currently in contact with the support of ptgrey.
> 
> Is it possible to reduce the buffer size? Then we should be able to reduce the effect of the delay.

Yes, the minimum buffer size is 3. But there was still a noticeable delay compared to the FirstOnly mode, as far as I could see.
But I'll test this again in more detail this evening.

g3force:
I made the relevant streaming parameters changeable in the UI and made 'OldestFirst' the default, which is stable at the moment.
When ssl-vision runs at full fps (not dropping frames), there is no noticeable latency, regardless of the streaming mode. 
I artificially reduced the processing frame rate by putting a usleep(30000) right before pCam->getNextImage(). This drops the frame rate to <30fps and there is a significant latency for all modes except FirstOnly.

I'd say the PR is ready, but default parameters might be changed if the problem with the SDK gets solved.

joydeep-b:
Okay, the usleep was just for the test, correct?

g3force:
> Okay, the usleep was just for the test, correct?

Yes, to simulate overload.

PR #127, opened 2019-02-24
Split camera images

### LINKED ISSUE #116: Split camera images and distribute them as a virtual source

g3force:
Resolves #116 

 * Add Distributor stack that captures an image and distributes it to the splitters
 * Add Splitter capture which cuts a part of the images from the distributor and applies the default vision stacks on the resulting image
 * Reduce the buffer size for capture to visualization transmission to reduce latency, always read the latest element from buffer
 * Add CLI parameter to set the number of camera threads

### DIFF
--- a/src/app/gui/mainwindow.cpp
+++ b/src/app/gui/mainwindow.cpp
@@ -21,7 +21,7 @@
 
 #include "mainwindow.h"
 
-MainWindow::MainWindow(bool start_capture, bool enforce_affinity)
+MainWindow::MainWindow(bool start_capture, bool enforce_affinity, int num_cameras)
 {
 
   affinity=0;
@@ -50,7 +50,7 @@ MainWindow::MainWindow(bool start_capture, bool enforce_affinity)
   //opt->parse();
 
   //load RoboCup SSL stack by default:
-  multi_stack= new MultiStackRoboCupSSL(opts);
+  multi_stack= new MultiStackRoboCupSSL(opts, num_cameras);
 
   VarExternal * stackvar;
   root->addChild(stackvar= new VarExternal((multi_stack->getSettingsFileName() + ".xml").c_str(),multi_stack->getName()));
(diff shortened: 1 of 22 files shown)


src/app/gui/mainwindow.cpp
```diff
@@ -21,7 +21,7 @@
 
 #include "mainwindow.h"
 
-MainWindow::MainWindow(bool start_capture, bool enforce_affinity)
+MainWindow::MainWindow(bool start_capture, bool enforce_affinity, int num_cameras)
```
joydeep-b: :+1: on making this a CLI parameter! :-)

src/app/stacks/multistack_robocup_ssl.cpp
```diff
@@ -84,14 +86,16 @@ MultiStackRoboCupSSL::MultiStackRoboCupSSL(RenderOptions * _opts, int cameras) :
...
+  createThreads(num_threads);
+  std::vector<CaptureSplitter*> captureSplitters;
+  captureSplitters.resize((unsigned long) num_normal_camera_threads);
+  for (unsigned int i = 0; i < num_normal_camera_threads;i++) {
     //NOTE: if modified to put different stacks in cameras, please
     //      update the plugin_colorcalib.cpp code to safely reallocate and copy their
     //      data instead of assuming that format and size is uniform across
     //      cameras -- added when LUTs became aware of other cameras (Zavesky, 2/16)
-    threads[i]->setFrameBuffer(new FrameBuffer(5));
+    threads[i]->setFrameBuffer(new FrameBuffer(3));
```
joydeep-b: Why 3?

src/app/stacks/multivisionstack.cpp
```diff
@@ -48,7 +48,7 @@ string MultiVisionStack::getSettingsFileName() {
 
 void MultiVisionStack::createThreads(int number) {
   for (int i=0;i<number;i++) {
-    threads.push_back(new CaptureThread(i));
+    threads.push_back(new CaptureThread(i%4));
```
joydeep-b: Should this be `num_cameras` instead of `4`?

src/app/gui/mainwindow.cpp
g3force: Have you read my comment on the old PR? https://github.com/RoboCup-SSL/ssl-vision/pull/117#discussion_r259627279

src/app/stacks/multistack_robocup_ssl.cpp
g3force: That's the minimum. With less, the application will block.

src/app/stacks/multivisionstack.cpp
g3force: Yeah, that should also work. Definitely looks better than the hard coded number :)

src/app/gui/mainwindow.cpp
joydeep-b: I see. Okay, that's fine for now.

Issue #128, opened 2019-02-26
Remove ifdefs for VDATA_NO_QT - QT is always required


g3force:
VarTypes allowed building without QT and that made it into ssl-vision with the initial commit.

In VarTypes, this has been removed in 2010: https://github.com/szi/vartypes/commit/d70feb7e30ee4432354fe42e9a09ae411d984de7

The `#ifndef VDATA_NO_QT` pollute the code and should thus be removed.

g3force:
that's actually quite easy:

```
find . -name "*.cpp" -exec unifdef -UVDATA_NO_QT {} -o {} \;
find . -name "*.h" -exec unifdef -UVDATA_NO_QT {} -o {} \;
```

PR #129, opened 2019-03-23
[#128] Remove VDATA_NO_QT ifdefs

g3force:
QT is mandatory for ssl-vision anyway, so no need to pollute code
with those ifdefs


g3force:
closes #128

g3force:
The changes were automatically done by unifdef (see #128).

PR #130, opened 2019-03-23
[#87] Add type enum to geometry proto for lines and arcs

### LINKED ISSUE #87: Add an enum-typed attribute to geometry message to identify lines and arcs
The current geometry message identifies lines and arcs by a string. The protocol is missing a concrete definition of each of the shapes. Teams have to look for certain names as implicitly defined in the source code to get dimensions of the penalty area.

I suggest to add an optional attribute to `SSL_FieldCicularArc` and `SSL_FieldLineSegment`:

```
enum SSL_FieldShapeType {
  CenterCircle = 0;
  TopTouchLine = 1;
  BottomTouchLine = 2;
  LeftGoalLine = 3;
  RightGoalLine = 4;
  HalfwayLine = 5;
  CenterLine = 6;
  LeftPenaltyStretch = 7;
  RightPenaltyStretch = 8;
  LeftFieldLeftPenaltyStrech = 9;
  LeftFieldRightPenaltyStrech = 10;
  RightFieldLeftPenaltyStrech = 11;
  RightFieldRightPenaltyStrech = 12;
}

message SSL_FieldLineSegment {
...
optional SSL_FieldShapeType = 5;
...
}

message SSL_FieldCicularArc {
...
optional SSL_FieldShapeType = 7;
...
}
```

g3force:
closes #87

### DIFF
--- a/src/client/main.cpp
+++ b/src/client/main.cpp
@@ -118,6 +118,8 @@ int main(int argc, char *argv[])
                 printf("  -boundary_width=%d (mm)\n",field.boundary_width());
                 printf("  -goal_width=%d (mm)\n",field.goal_width());
                 printf("  -goal_depth=%d (mm)\n",field.goal_depth());
+                printf("  -penalty_area_depth=%d (mm)\n",field.penalty_area_depth());
+                printf("  -penalty_area_width=%d (mm)\n",field.penalty_area_width());
                 printf("  -field_lines_size=%d\n",field.field_lines_size());
                 printf("  -field_arcs_size=%d\n",field.field_arcs_size());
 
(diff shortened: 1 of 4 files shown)


g3force:
I started this one last year. Time to finish it up.

joydeep-b:
I went over this, and it seems fine, just want to confirm the use case - is the idea to specify what part of the field each line is supposed to represent?
@g3force

joydeep-b:
Ah, missed the conversation in the Issue. Sounds good.

joydeep-b:
@g3force Feel free to merge when ready.

g3force:
yeah. Right now, all the lines and arcs have standardized names, but they are encoded as strings. To find the field dimensions, you have to expect lines with certain names.

I think, the reason why I still not merged this is, that I'd like to also have some parameters directly in the protobuf structure, like field length, etc, so that teams do not need extract them from the lines.

I'll have a look at it later.

joydeep-b:
We can have a redundant `length` field for the lines, to make it easy.

g3force:
I mixed things up. I thought there is no field length and width in the protobuf, but it is. The lines should be fine as is. I'll have a more detailed look later and get this merged.

g3force:
I found, what I wanted to add: The penalty area dimensions where only implicitly encoded into the lines. I added dedicated fields next to the field dimensions, so teams do not have to extract the dimensions from the lines.

PR #133, opened 2019-05-22
Performance measurement

g3force:
I added an option to print several timings of image processing. I think this will also be useful for end users if the capture rate is low, to see where the bottleneck is.

The steady clock should have nearly no impact on the performance.

### DIFF
--- a/src/app/stacks/visionstack.cpp
+++ b/src/app/stacks/visionstack.cpp
@@ -19,10 +19,13 @@
 */
 //========================================================================
 #include "visionstack.h"
+#include <iomanip>
 
 VisionStack::VisionStack(RenderOptions * _opts) {
   opts=_opts;
   settings=new VarList("Global");
+  _v_print_timings = new VarBool("print stack timings", false);
+  settings->addChild(_v_print_timings);
 }
 
 VisionStack::~VisionStack() {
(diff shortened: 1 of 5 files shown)


src/app/stacks/visionstack.cpp
```diff
@@ -34,10 +39,26 @@ VarList * VisionStack::getSettings() {
 }
 
 void VisionStack::process(FrameData * data) {
-  for (auto p : stack) {
-    p->lock();
-    p->process(data,opts);
-    p->unlock();
+  if(_v_print_timings->getBool()) {
```
joydeep-b: This looks like it will print the stats for every frame, for every camera. That is a lot of stdout processing. I recommend the numbers be accumulated and averaged over (for example) 30 frames, and printed only every 0.5 seconds.

g3force: my workflow was like this: starting ssl-vision, selecting the desired camera and activate the "print timings" flag for a short time. Then I had a look at the console to check the timings.

Having this flag active for a longer time will probably never be useful.

Accumulating results could be nice, but it would also increase complexity in the code and might hide spikes in the processing times. So I'd rather leave it like it is now.

joydeep-b: That sounds fine, as long as we disclose this: that it is not intended to be used during actual games, or running continuously with the timings, and that it is most useful only running in single-camera mode for a test period.

g3force: ok. Do you mean to document this somewhere? I'm not sure, where I should do this. The flag is false by default, so it must be switched on actively per camera.
If this is fine for you, I would merge the PR?

joydeep-b:
Perhaps as a comment in the code.
Yes, good to merge.

PR #134, opened 2019-05-24
Threaded threshold

g3force:
Allow performing the color threshold on multiple threads to reduce processing time on large images.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,6 +251,7 @@ qt4_wrap_cpp (MOC_SRCS
 	src/app/plugins/plugin_legacypublishgeometry.h
 	src/app/plugins/visionplugin.h
 	src/app/plugins/plugin_colorcalib.h
+	src/app/plugins/plugin_colorthreshold.h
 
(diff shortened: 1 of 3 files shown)


g3force:
I measured the processing time on a set of raw images from the new FLIR camera provided by RoboDragons. Most significant numbers (time per frame):
raw -> rgb conversion: 1.3ms
thresholding: 6.5ms
runlength encode: 2.4ms
visualization (after network publish): 1.3ms

Background for those changes: We noticed that it might be a problem for some teams if the camera center is the same on multiple virtual cameras (as it is for the split image solution).
Threshold and runlength encode are the most time consuming operations and they can easily be parallelized by splitting the image into equal sized rows (no copying required).

I can share performance comparisons, once PR #133 is merged.

After this PR is merged, I would consider making the runlength encode parallelized too, but in this case with a small overlap.

g3force:
Based on the fps, the change seems to be effective. On my laptop with Quad-Core i7 I get:
0 threads -> 61 fps
2 threads -> 73 fps
4 threads -> 79 fps
8 threads -> 82 fps

PR #135, opened 2019-05-27
Optimize color thresholding by using SIMD instructions

g3force:


joydeep-b:
Can you share the speedup numbers, @g3force ?
This should be hidden behind an SSE detection flag to ensure it is compiled if (and only if) the target supports SSE3. 

Potential solution:
https://stackoverflow.com/questions/28939652/how-to-detect-sse-sse2-avx-avx2-avx-512-avx-128-fma-kcvi-availability-at-compile

g3force:
https://docs.google.com/spreadsheets/d/1NSrU4v9CXz7-20ZYo8ViBtGx1fmVF3BF-aC9ySRO4Ss/edit#gid=0

g3force:
The numbers are from my laptop. I used the fromfile capture module to read raw images.
Speedup without threads is 36%, with 8 threads 10%

Btw, most of the code is from @andre-ryll

g3force:
I use the `__AVX2__`  detection flag now.

andre-ryll:
I tested this on the designated NUC with the real camera and with visualisation turned off this gets to max fps (73) easily. Around ~40% total CPU usage.

g3force:
any further concerns? Otherwise, I'll merge it.

PR #136, opened 2019-05-31
Allow disabling most capture modules + Build different configurations

g3force:
I added defines for excluding v4l and splitter/distributor and unified the configuration in CMakeLists.txt so that all camera modules can be switched on/off with the USE_* cmake cached variables.

Firewire is still on by default, but the other modules have to be explicitly activated. If activated, the build will also fail if the SDK is not found.

There are now four build configs for CI: default, with v4l, with splitter, with qt5. The camera modules require third party SDK that can not easily be installed automatically in CI... :/


CMakeLists.txt
```diff
@@ -3,8 +3,12 @@ set(OpenGL_GL_PREFERENCE LEGACY)
 set(USE_QT5 FALSE CACHE BOOL "Use Qt5 instead of Qt4")
 set(USE_NEURAL_CALIBRATION FALSE CACHE BOOL "Enable neural network calibration (requires OpenCV 2, not compatible with later versions)")
 set(USE_DC1394 TRUE CACHE BOOL "Compile with DC1394 driver (firewire cameras)")
-set(USE_SPINNAKER FALSE CACHE BOOL "Compile with Spinnaker driver (FLIR USB3 cameras)")
+set(USE_SPINNAKER FALSE CACHE BOOL "Compile with Spinnaker driver (FLIR cameras)")
```
joydeep-b: I think the Spinnaker option should be true by default, since it will be the official version at the competition from now on.

README.md
```diff
@@ -41,10 +41,29 @@ Multiple cameras are supported:
  * Basic usb camera support via the [Video for Linux (V4L)](http://linuxtv.org/downloads/v4l-dvb-apis/) drivers
  * Matrix-Vision BlueFox (USB 2.0) and BlueFox3 (USB 3.0) cameras via [mvIMPACT Acquire SDK](http://www.matrix-vision.com/software-drivers-en.html)
  * Basler cameras via the [Pylon Software Suite](https://www.baslerweb.com/en/products/software/basler-pylon-camera-software-suite/)
- * FLIR cameras via the [SPINNAKER SDK](https://www.ptgrey.com/support/downloads)
+ * FLIR cameras via the [SPINNAKER and FLYCAP SDK](https://www.flir.com/support-center/iis/machine-vision/downloads/spinnaker-sdk-flycapture-and-firmware-download/)
```
joydeep-b: Probably good to point out that Flycap is not compatible with Ubuntu 18.04, and Spinnaker is officially only compatible with only 16.04 and 18.04.

CMakeLists.txt
g3force: if it is on by default, you have to either install the SDK or disable it explicitly. Since most teams will still not use it in their labs, I would rather keep it off by default.
One option would be, to use it only, if the SDK was found, like it has been before. But I prefer the explicit way, so that the build will fail early, if it switched on, but the SDK was not found.

README.md
g3force: Actually, flywap is compatible with 18.04. At least there is a package from April, 4th 2019.

I think the information about the compatibility should rather come from the download side, as it can quickly get outdated.

joydeep-b: Ah, that's new, thanks for checking! Yes, given that these are changing often, that sounds good.

CMakeLists.txt
joydeep-b: Okay.

PR #137, opened 2019-06-01
Add max_regions and max_runs to config tree, increase default values

g3force:
With increased image sizes, the maximum values may get exceeded.

### DIFF
--- a/src/shared/cmvision/cmvision_region.h
+++ b/src/shared/cmvision/cmvision_region.h
@@ -28,8 +28,6 @@
 #include "cmvision_threshold.h"
 #include "lut3d.h"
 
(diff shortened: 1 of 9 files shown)


src/shared/cmvision/cmvision_region.h
```diff
@@ -28,8 +28,6 @@
 #include "cmvision_threshold.h"
 #include "lut3d.h"
 
-#define CMV_DEFAULT_MAX_RUNS 100000
```
joydeep-b: Was this not used anywhere else? It doesnt' show up in the diff for the PR.

g3force: no, was not used. The numbers were hard-coded before.

PR #138, opened 2019-06-13
Init color calibration

g3force:
This adds the initial color calibrator that was created by Mark two years ago.

I changed the way how calibration points a set. There is a new widget that allows resetting the lut and adding samples for different colors.


g3force:
oh, yeah. I was a bit lazy. I'll move those changes to a new PR.

PR #141, opened 2019-06-15
Improved some default parameters

g3force:


g3force:
#139 needs to be merged first

src/app/plugins/plugin_detect_balls.h
```diff
@@ -74,11 +74,11 @@ friend class PluginDetectBalls;
 
   _settings=new VarList("Ball Detection");
 
-  _settings->addChild(_max_balls = new VarInt("Max Ball Count",10));
+  _settings->addChild(_max_balls = new VarInt("Max Ball Count",20));
   _settings->addChild(_color_label = new VarString("Ball Color","Orange"));
 
   _settings->addChild(_filter_general = new VarList("Ball Properties"));
-    _filter_general->addChild(_ball_z_height = new VarDouble("Ball Z-Height", 30.0));
+    _filter_general->addChild(_ball_z_height = new VarDouble("Ball Z-Height", 21.5));
```
g3force: I'm still not sure if the 30mm were intended.
The ball has a radius of 21.5mm. When looking from above, I see the full shape of the ball, which is at a height of one radius.

g3force: I've experimented with this value on my field and it turned out that 30mm actually seems to be good. I still do not fully understand, why, though.

joydeep-b: Yes, we have consistently observed the same, which is why it is so. Can you add a comment to the effect? Let's prevent people in the future try to set it back. ;-)
I suspect it is partially due to the ball sinking into the carpet, and partially due to bias in the geometry calibrator.

src/shared/util/camera_calibration.cpp
```diff
@@ -745,12 +745,9 @@ CameraParameters::AdditionalCalibrationInformation::
     }
 
   initial_distortion = new VarDouble("initial distortion", 1.0);
-  camera_height = new VarDouble("camera height", 4000.0);
-  line_search_corridor_width = new VarDouble("line search corridor width",
-                                             280.0);
+  line_search_corridor_width = new VarDouble("line search corridor width", 280.0);
```
joydeep-b: This change might conflict with the cleanup PR #139 
Recommend you merge #139, then pull into this branch first.

src/app/plugins/plugin_detect_balls.cpp
```diff
@@ -248,7 +248,9 @@ ProcessResult PluginDetectBalls::process ( FrameData * data, RenderOptions * opt
             for (int r = 0; r < robots_n; r++) {
               const SSL_DetectionRobot & robot = robots->Get(r);
               if (robot.confidence() > 0.0) {
-                if ((sq((double)(robot.x())-(double)(field_pos.x)) + sq((double)(robot.y())-(double)(field_pos.y))) < near_robot_dist_sq) {
+                vector3d field_on_bot_pos_3d;
+                camera_parameters.image2field ( field_on_bot_pos_3d, pixel_pos, robot.height());
```
joydeep-b: Why is this necessary? It is already done in lines 217-220.

src/shared/util/camera_calibration.cpp
g3force: I already noted this in a comment above ;) I based this branch on the cleanup branch, so that there will be no conflict ;)

src/app/plugins/plugin_detect_balls.cpp
g3force: the difference is, that here I transform the blob to robot height.
Because if e.g. a magenta blob is falsely detected as orange und thus inside the bot pattern, it is at robot height. The filter did not behave correctly before.

src/app/plugins/plugin_detect_balls.h
g3force: done.

src/app/plugins/plugin_detect_balls.cpp
joydeep-b: Ah, I see. Yes, makes sense.

PR #150, opened 2019-10-03
Hotfix/integer division

### LINKED ISSUE #149: Basler camera timing integer division
There is an issue in  src/shared/capture/capture_basler.cpp, line 190,  (integer division), causing the basler camera timings to only be accurate up to a second. We (Roboteam Twente) fixed it downstream but we have a lot of changes on our fork so it's probably best if fixed separately here. It should be as simple as adding .0 behind the division denumerator.

rolfvdhulst:
Fixes an integer division bug which makes timings only accurate up to 1 second.
Has been tested with the actual camera´s to verify the fix.

fixes #149


g3force:
You've replaced the Makefile by some large generated one ;) And there are some unrelated formatting changes btw.
Wouldn't it be easier to just make the single change, instead of reverting all the unrelated changes? ;)

rolfvdhulst:
I made a new pr from my personal fork. We are running only our fork with some edits (of which I don't know the exact nature) on our vision computers, so I had no easy way to revert changes whilst also testing with the real camera's. I made a PR with the isolated changes from my own fork now (see #157 ), which should just be fine in itself.

g3force:
Ok. Would still be interesting to compare your diffs with the master and see, if there are other fixes that are relevant for other users.

rolfvdhulst:
As far as I am aware there is nothing, we're a bit behind master and have some hooks for our own custom vision system that uses ArUco tags, but that is not performing well enough at the moment to be relevant for others (we run into similar problems as mentioned in the april tags pr, mostly camera distortion and motion blur issues with certain camera's, and simply performance issues with some computations being too slow).

PR #151, opened 2019-10-25
Apriltags

rhololkeolke:
Adds a new optional apriltag robot detector plugin. See the README updates for more information about how to compile, run, and configure this detector.


rhololkeolke:
Can we merge this in before master diverges too much?

Other than it being a bit slow for high res images, everything seems to work for those that have tried it. And the speed-ups for high res images can be merged in a different request.

src/app/plugins/plugin_detect_robots.cpp
```diff
@@ -126,14 +126,14 @@ ProcessResult PluginDetectRobots::process(FrameData * data, RenderOptions * opti
       color_id=color_id_blue;
       team=global_team_selector_blue->getSelectedTeam();
       num_robots=global_team_selector_blue->getNumberRobots();
-      detection_frame->clear_robots_blue();
+      // detection_frame->clear_robots_blue();
```
g3force: Why is this commented out? Can this be removed safely?

src/app/plugins/plugin_apriltag.cpp
```diff
@@ -0,0 +1,303 @@
...
+                        p_bottom_left_field.x - p_top_left_field.x);
+
+      SSL_DetectionRobot *robot_detection = nullptr;
+      if (tag_team == Team::Blue) {
+        robot_detection = detection_frame->add_robots_blue();
+      } else {
+        robot_detection = detection_frame->add_robots_yellow();
+      }
+      robot_detection->set_confidence(
+          det->decision_margin); // TODO(dschwab): What value should I put here?
```
g3force: is this still a TODO? Can this be resolved?

src/shared/cmpattern/cmpattern_teamdetector.cpp
```diff
@@ -181,7 +181,7 @@ TeamDetector::~TeamDetector()
 void TeamDetector::update(::google::protobuf::RepeatedPtrField< ::SSL_DetectionRobot >* robots, int team_color_id, int max_robots, const Image<raw8> * image, CMVision::ColorRegionList * colorlist, CMVision::RegionTree & reg_tree) {
   color_id_team=team_color_id;
   _max_robots=max_robots;
-  robots->Clear();
+  // robots->Clear();
```
g3force: same here: Can this be removed?

g3force:
Automatic build for apriltag added to circleci config. It checks out the master branch of the apriltag library, so the build is not reproducible, but we will find issues with updates in the apriltag lib earlier.

rdesc:
The updated README.md should explicity say the april tags repo needs to be cloned and installed. Also I had to update my LD_LIBRARY_PATH env variable to include the path where the libapriltag.so files were installed, might be worth mentioning in the docs as well.

PR #158, opened 2019-12-01
Apply a mask to the thresholded image to ignore the surroundings

g3force:


g3force:
Some things I noticed so far:
* The widget should not be the upper most
* The widget should have some explaination of what to do
* The naming in the UI could be improved
* The fps drops significantly when enabling visualization of the mask

But it is functional and works at 70fps on the high-res camera

g3force:
I've removed the empty widget and instead put a toggle button into the lut widget toolbar. If pressed, the mask can be drawn.
I've also renamed 'mask' to 'image mask'.
The fps will still drop, if visualizing the mask (blackout the borders). I think it is ok to leave it as is.

With this, this PR is ready to be reviewed and tested.

src/app/gui/glLUTwidget.cpp
```diff
@@ -464,7 +464,7 @@ GLLUTWidget::GLLUTWidget(LUTChannelMode mode, QWidget *parent) : QGLWidget(paren
   addAction(actionUndo);
 
   actionRedo = new QAction(this);
-  actionUndo->setEnabled(false);
+  actionRedo->setEnabled(false);
```
joydeep-b: Is this a bug fix?

src/app/gui/maskwidget.h
```diff
@@ -0,0 +1,17 @@
+#ifndef MASKWIDGET_H
+#define MASKWIDGET_H
+#include <QWidget>
+
+/*!
+  \file    cameracalibwidget.h
+  \brief   A GUI for setting up a mask on the image
+  \author  Rui Silva, (C) 2017
+*/
+class MaskWidget : public QWidget {
```
joydeep-b: This class does not seem to be actually used anywhere, or did I miss something?

src/shared/util/convex_hull.cpp
```diff
@@ -0,0 +1,87 @@
...
+*/
+Points computeConvexHullFromPoints(Points &P) {
+  const int n = P.size();
+  if (n == 1) return P;
+  
+  Points R(2 * n);
+  std::sort(P.begin(), P.end(), lexicographicLess);
+
+  int k = 0;
+  // Lower hull
```
joydeep-b: How many points are usually used? If it is large, I suggest using one of the O(n log(n)) algorithms like Graham Scan: https://en.wikipedia.org/wiki/Convex_hull_algorithms
Maybe this can be tracked as an enhancement feature for later optimization.

joydeep-b:
@g3force Thanks for this PR. It should help avoid the problem of SSL-Vision trying to get blobs from non-field objects and choking on them.
I skimmed through the PR, and it seems good, I added a few questions. 
I have not run this PR, so I recommend asking someone else to try it out before merging, since it affects core parts of ssl-vision, the LUT, etc.

src/app/gui/glLUTwidget.cpp
g3force: yes, looks like a copy paste error. actionUndo was disabled before already.

src/app/gui/maskwidget.h
g3force: good catch, forgot to remove it.

src/shared/util/convex_hull.cpp
g3force: depends on the camera distortion. With the new FLIR cameras, 4 points are already sufficient.
@rhololkeolke how many points have you used?

rhololkeolke: It depends on how distorted your cameras are and how much noise you have around the field. I think we were using about 8 points for some of the cameras in our previous lab setup. 

I had put a version of this on the field computers at last years competition and I think it was between 4 and 8 points. I wasn't super exact with the boundaries. Just used it to cut out the crowds walking by and robots sitting on the tables.

rttwente:
Tested this with our Basler cameras, and found no issues

src/app/plugins/plugin_visualize.cpp
```diff
@@ -101,12 +109,27 @@ void PluginVisualize::DrawCameraImage(
...
+    rgb * vis_ptr = vis_frame->data.getPixelData();
+    rgb color;
+
+    _image_mask.lock();
+    for (unsigned int i = 0; i < n; i++) {
+      vis_ptr[i].r &= _image_mask.getMask().getPixel(i).getIntensity();
+      vis_ptr[i].g &= _image_mask.getMask().getPixel(i).getIntensity();
+      vis_ptr[i].b &= _image_mask.getMask().getPixel(i).getIntensity();
+    }
+    _image_mask.unlock();
```
rhololkeolke: I imagine the slow down when visualizing is due to this block. We should be able to replace this with the opencv mask functions or at a minimum memset/memcpy. But I think opencv will take advantage of vectorizing loops for certain calls, so we'll probably have better luck with that.

Its not a blocker for merging this, since the core functionality doesn't rely on this, but maybe we should at least add a TODO note for future people looking at this section until it is fixed.

g3force:
I've put the button and a description below the LUT to make it more intuitive
image

g3force:
There is one weird bug: The description label text is turned white when the LUT widget has focus. If some other part of the UI gets focus, the text is turned black again. And if the Edit Mask button is clicked, the text also turned black again.

src/app/plugins/plugin_visualize.cpp
g3force: In any case, it will add to the overall processing time for visualization, which is already a lot. I think there is no real benefit in having the black background compared to the hull.

I've removed this visualization completely (and kept the hull). If you disagree, let me know and I revert it.

src/shared/util/convex_hull.cpp
g3force: so it seems there is currently no need to optimize it. the convex hull is used to generate the masked image on changes only anyway, so this should not have an influence on the image processing time.

g3force:
I've moved the UI to a separate widget again, as I could not fix the bug mentioned above...

g3force:
I've also optimized the code for applying the mask by using the mask in the threshold loop already instead of looping over the whole image again.

rhololkeolke:
I just tested the latest version and it all seems to be working for me. Performance is fine (but I'm using relatively low-res cameras).

Regarding the UI, while it takes up more space, I think having it as the separate widget with instructions is more helpful. And the space becomes less of an issue once #164 is merged.

PR #162, opened 2020-01-01
New protocol for dedicated tracking software

g3force:
Draft for a new protocol for dedicated tracking software

### DIFF
--- a/src/shared/proto/messages_robocup_ssl_detection_tracked.proto
+++ b/src/shared/proto/messages_robocup_ssl_detection_tracked.proto
@@ -66,22 +66,18 @@ message TrackedRobot {
     // The exact implementation depends on the source software
     optional float quality = 2;
 
-    // The x-position [m] in the ssl-vision coordinate system
-    required float pos_x = 3;
-    // The y-position [m] in the ssl-vision coordinate system
-    required float pos_y = 4;
+    // The position [m] in the ssl-vision coordinate system
+    required Vector2 pos = 3;
     // The orientation [rad] in the ssl-vision coordinate system
-    required float orientation = 5;
+    required float orientation = 4;
 
-    // The x-velocity [m/s] in the ssl-vision coordinate system
-    optional float vel_x = 6;
-    // The y-velocity [m/s] in the ssl-vision coordinate system
-    optional float vel_y = 7;
+    // The velocity [m/s] in the ssl-vision coordinate system
+    optional Vector2 vel = 5;
     // The angular velocity [rad/s] in the ssl-vision coordinate system
-    optional float vel_angular = 8;
+    optional float vel_angular = 6;
 
     // The robot height
-    required float height = 9;
+    required float height = 7;
 }
 
 // A frame that contains all currently tracked objects on the field on all cameras
--- a/src/shared/proto/messages_robocup_ssl_detection_tracked.proto
+++ b/src/shared/proto/messages_robocup_ssl_detection_tracked.proto
@@ -1,10 +1,12 @@
 syntax = "proto2";
 
+// A vector with two dimensions
 message Vector2 {
     required float x = 1;
     required float y = 2;
 }
 
+// A vector with three dimensions
 message Vector3 {
     required float x = 1;
     required float y = 2;
@@ -25,34 +27,18 @@ message RobotId {
     required TeamColor team_color = 2;
 }
 
-enum ChipKickDetectionSupport {
-    UNSUPPORTED = 0;
-    FIRST_TOUCHDOWN_ONLY = 1;
-    ALL_TOUCHDOWNS = 2;
-}
-
 // A single tracked ball
 message TrackedBall {
-    // The quality of the detection
-    // A value between 0 (not visible) and 1 (visible)
-    // The exact implementation depends on the source software
-    optional float quality = 1;
-
     // The position (x, y, height) [m] in the ssl-vision coordinate system
-    required Vector3 pos = 2;
+    required Vector3 pos = 1;
 
     // The velocity [m/s] in the ssl-vision coordinate system
-    optional Vector3 vel = 3;
+    optional Vector3 vel = 2;
 
-    // The level of chip kick detection support by the source implementation
-    required ChipKickDetectionSupport chip_kick_detection_support = 4;
-
-    // The touch down location(s), if the ball is flying.
-    // Depending on the source implementation, the next or all touch down locations may be provided
-    // ChipKickDetectionSupport#UNSUPPORTED -> there are no valid touch down locations
-    // ChipKickDetectionSupport#FIRST_TOUCHDOWN_ONLY -> there is only one touch down location, although the ball may have more
-    // ChipKickDetectionSupport#ALL_TOUCHDOWNS -> all touch down locations are included
-    repeated Vector2 touch_down_locations = 5;
+    // The visibility of the ball
+    // A value between 0 (not visible) and 1 (visible)
+    // The exact implementation depends on the source software
+    optional float visibility = 3;
 }
 
 // A ball kicked by a robot, including predictions when the ball will come to a stop
@@ -62,41 +48,38 @@ message KickedBall {
     // The initial velocity [m/s] with which the ball was kicked
     required Vector3 vel = 2;
     // The unix timestamp [s] when the kick was performed
-    required double timestamp = 3;
+    required double start_timestamp = 3;
 
-    // The predicted time [s] that ball ball needs to come to a stop
-    optional float stop_time = 4;
+    // The predicted unix timestamp [s] when the ball comes to a stop
+    optional double stop_timestamp = 4;
     // The predicted position [m] at which the ball will come to a stop
     optional Vector2 stop_pos = 5;
 
     // The robot that kicked the ball
     optional RobotId robot_id = 6;
-
-    // A flag to indicate if the ball was chipped initially
-    optional bool chipped = 7;
 }
 
 // A single tracked robot
 message TrackedRobot {
     required RobotId robot_id = 1;
 
-    // The quality of the detection
-    // A value between 0 (not visible) and 1 (visible)
-    // The exact implementation depends on the source software
-    optional float quality = 2;
-
     // The position [m] in the ssl-vision coordinate system
-    required Vector2 pos = 3;
+    required Vector2 pos = 2;
     // The orientation [rad] in the ssl-vision coordinate system
-    required float orientation = 4;
+    required float orientation = 3;
 
     // The velocity [m/s] in the ssl-vision coordinate system
-    optional Vector2 vel = 5;
+    optional Vector2 vel = 4;
     // The angular velocity [rad/s] in the ssl-vision coordinate system
-    optional float vel_angular = 6;
+    optional float vel_angular = 5;
 
-    // The robot height
-    required float height = 7;
+    // The robot height [m]
+    required float height = 6;
+
+    // The visibility of the robot
+    // A value between 0 (not visible) and 1 (visible)
+    // The exact implementation depends on the source software
+    optional float visibility = 7;
 }
 
 // A frame that contains all currently tracked objects on the field on all cameras
--- a/src/shared/proto/messages_robocup_ssl_detection_tracked.proto
+++ b/src/shared/proto/messages_robocup_ssl_detection_tracked.proto
@@ -37,6 +37,10 @@ message TrackedBall {
 
     // The velocity [m/s] in the ssl-vision coordinate system
     optional Vector3 vel = 3;
+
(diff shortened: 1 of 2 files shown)


src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,87 @@
...
+    optional float quality = 1;
+
+    // The position (x, y, height) [m] in the ssl-vision coordinate system
+    required Vector3 pos = 2;
+
+    // The velocity [m/s] in the ssl-vision coordinate system
+    optional Vector3 vel = 3;
+
+    // The initial (kick) speed [m/s], if the ball is moving
+    optional Vector3 initial_speed = 4;
```
joydeep-b: The definition of "initial" is not clear. It seems like it should be associated with an initial location and time.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,87 @@
...
+    optional float vel_y = 8;
+    // The angular velocity [rad/s] in the ssl-vision coordinate system
+    optional float vel_angular = 9;
+
+    // The robot height
+    required float height = 10;
+}
+
+// A frame that contains all currently tracked objects on the field
+message TrackedFrame {
```
joydeep-b: Is this for the entire field, by merging across all cameras, or for a single camera? Some clarification of the intent in the comments would help.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
g3force: true, I have declared a whole new message for a 'KickedBall' now.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
g3force: I took it for granted, that there is no separation between different cameras, but I added it to the comment now to make this clear.
The whole idea of the protocol is, that the receiver does not need to care about cameras at all ;)

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,106 @@
...
+message TrackedRobot {
+    required RobotId robot_id = 1;
+
+    // The quality of the detection
+    // A value between 0 (not visible) and 1 (visible)
+    // The exact implementation depends on the source software
+    optional float quality = 2;
+
+    // The x-position [m] in the ssl-vision coordinate system
+    required float pos_x = 3;
```
andreaswendler: Why not use the Vector messages from above here?

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,106 @@
...
+
+    // The position (x, y, height) [m] in the ssl-vision coordinate system
+    required Vector3 pos = 2;
+
+    // The velocity [m/s] in the ssl-vision coordinate system
+    optional Vector3 vel = 3;
+}
+
+// A ball kicked by a robot, including predictions when the ball will come to a stop
+message KickedBall {
```
andreaswendler: As we at ER-Force will implement this protocol, I can already say that we currently do not have the required information to fill this message. While it may be produced as intermediate steps in the tracking computation, we do not propagate it to other parts of our framework. That is why we would likely have to omit this message.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
g3force: good catch. I forgot it there after introducing the vector messages. Fixed now.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
g3force: That's fine, the kicked_ball field is optional in the frame. But I added a note that explicitly says that some implementations will not set it at all.
Or could you set some of the fields in this message? In this case, are the optional and required attributes ok?

andreaswendler: Without changes to our internal messages, I could currently not send out any fields of this message. The position, velocity and timestamp is internal to the tracking and the stop position and time are not part of our tracking but computed in the typescript/lua strategy code, to which the planned CLI for this will have no access.
Both could be changed, but currently we only have the next touchdown position of the ball (when it starts the next bounce) and if it is currently bouncing as additional data.

g3force: next touch down would indeed be interesting, I'll add this as well.

g3force: added. Is it sufficient to encode "currently bouncing" as an empty list of touch down locations?
We provide all touch down locations, that's why I defined a list, but it's fine to just add one location.

andreaswendler: Should be fine, yes

joydeep-b: There seems to be some ambiguity in how to interpret this message as is: The touch down locations may 1) be absent entirely, 2) contain only the first touch down location, or 3) contain all subsequent locations (or some subset thereof). 
It would be good to have an enum to indicate which of these options apply. For example, using something like:
```
enum ChipKickDetectionSupport {
  UNSUPPORTED = 0;
  FIRST_TOUCHDOWN = 1;
  MULTIPLE_TOUCHDOWNS = 2;
}
```

g3force: good point. I've added an Enum similar to your suggestion.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,122 @@
...
+
+    // The level of chip kick detection support by the source implementation
+    required ChipKickDetectionSupport chip_kick_detection_support = 4;
+
+    // The touch down location(s), if the ball is flying.
+    // Depending on the source implementation, the next or all touch down locations may be provided
+    // ChipKickDetectionSupport#UNSUPPORTED -> there are no valid touch down locations
+    // ChipKickDetectionSupport#FIRST_TOUCHDOWN_ONLY -> there is only one touch down location, although the ball may have more
+    // ChipKickDetectionSupport#ALL_TOUCHDOWNS -> all touch down locations are included
+    repeated Vector2 touch_down_locations = 5;
```
andre-ryll: I would recommend to remove this field completely. The first touchdown location is implicit in the ball position and velocity as it fully describes the ball state to compute a touchdown.
Further touchdown locations have very low confidence. Bounce behaviour depends a lot on (not observable) ball spin. It also differes slightly depending on touchdown location on the field (due to construction differences).
I think this field is rather confusing for the intended users (new teams) and would drop it completely.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,122 @@
...
+    optional float quality = 1;
+
+    // The position (x, y, height) [m] in the ssl-vision coordinate system
+    required Vector3 pos = 2;
+
+    // The velocity [m/s] in the ssl-vision coordinate system
+    optional Vector3 vel = 3;
+
+    // The level of chip kick detection support by the source implementation
+    required ChipKickDetectionSupport chip_kick_detection_support = 4;
```
andre-ryll: Would recommend to move this to a general "capabilities" field in the TrackerWrapperPacket to indicate what the sending software supports. No need to have it in every TrackedBall.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,122 @@
...
+message KickedBall {
+    // The initial position [m] from which the ball was kicked
+    required Vector3 pos = 1;
+    // The initial velocity [m/s] with which the ball was kicked
+    required Vector3 vel = 2;
+    // The unix timestamp [s] when the kick was performed
+    required double timestamp = 3;
+
+    // The predicted time [s] that ball ball needs to come to a stop
+    optional float stop_time = 4;
```
andre-ryll: It is not fully clear if this is a timestamp or time until stop. I would actually recommend a timestamp so it does not change constantly.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,122 @@
...
+    // The predicted position [m] at which the ball will come to a stop
+    optional Vector2 stop_pos = 5;
+
+    // The robot that kicked the ball
+    optional RobotId robot_id = 6;
+
+    // A flag to indicate if the ball was chipped initially
+    optional bool chipped = 7;
```
andre-ryll: This is redundant and can be removed. An initial Z velocity != 0 indicates a chip kick.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,122 @@
...
+    // The orientation [rad] in the ssl-vision coordinate system
+    required float orientation = 4;
+
+    // The velocity [m/s] in the ssl-vision coordinate system
+    optional Vector2 vel = 5;
+    // The angular velocity [rad/s] in the ssl-vision coordinate system
+    optional float vel_angular = 6;
+
+    // The robot height
+    required float height = 7;
```
andre-ryll: Please specify a unit here.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,122 @@
...
+    FIRST_TOUCHDOWN_ONLY = 1;
+    ALL_TOUCHDOWNS = 2;
+}
+
+// A single tracked ball
+message TrackedBall {
+    // The quality of the detection
+    // A value between 0 (not visible) and 1 (visible)
+    // The exact implementation depends on the source software
+    optional float quality = 1;
```
andre-ryll: If this is an indicator for visibility I would name it like that.
On a next level it would actually make sense to add a covariance matrix to have a full picture of uncertainties. But I would consider this optional for this PR.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
g3force: agreed

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
g3force: without the touch down locations, this is not needed anymore. If we need this later, I like the idea of the capability list.

g3force: Actually, there are some interesting capabilities. I've added them like you suggested.

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,120 @@
...
+    // The visibility of the ball
+    // A value between 0 (not visible) and 1 (visible)
+    // The exact implementation depends on the source software
+    optional float visibility = 3;
+}
+
+// A ball kicked by a robot, including predictions when the ball will come to a stop
+message KickedBall {
+    // The initial position [m] from which the ball was kicked
+    required Vector3 pos = 1;
```
g3force: Is it safe to assume that balls are always kicked from the ground?

In this case, I would make this a Vector2

src/shared/proto/messages_robocup_ssl_detection_tracked.proto
```diff
@@ -0,0 +1,120 @@
...
+
+// A frame that contains all currently tracked objects on the field on all cameras
+message TrackedFrame {
+    // A monotonous increasing frame counter
+    required uint32 frame_number = 1;
+    // The unix timestamp in [s] of the data
+    // If timestamp is larger than timestamp_captured, the source has applied a prediction already
+    required double timestamp = 2;
+    // The unix timestamp in [s] of the source image as reported by ssl-vision
+    required double timestamp_captured = 3;
```
g3force: It is actually quite difficult to determine the captured time, because the frame contains data from multiple cameras with different captures times, so I suggest to remove this field.

g3force:
I removed the robot height.

Turned out the robot height is not that important, except maybe for the tracker itself. For higher level code, the maximum allowed height is sufficient due to accuracy of the detectable ball height.

g3force:
Ready to be merged. I've created a section on the website: https://ssl.robocup.org/league-software/

The protocol is already implemented in the TIGERs autoRef, in the Go logging tools and there is a test client in the go-tools.

PR #163, opened 2020-01-09
Camera calibration with OpenCV and checkerboard

g3force:
Based on the work of @rhololkeolke I've implemented the openCV chessboard calibration with intrinsic and extrinsic models.

So far, a calibration of the intrinsics with a chessboard can be done through a new widget. I've tested it with the original SSL FLIR cameras and a DIN A4 7x9 chessboard. The RMS error is shown and the result is saved.
Additionally, there is a button to calibrate the extrinsics based on the four field corner markers.

If enabled in the config tree (Camera Calibrator), the model is used for field2Image and image2Field, but the code is not yet optimized, especially the image2field where several matrix inversions are done.

The UI widget also still needs more structure.

I'll continue work on this, but if anyone wants to start testing this branch, you are welcome.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -22,8 +22,8 @@ AlwaysBreakAfterDefinitionReturnType: None
 AlwaysBreakAfterReturnType: None
 AlwaysBreakBeforeMultilineStrings: true
 AlwaysBreakTemplateDeclarations: Yes
-BinPackArguments: true
-BinPackParameters: true
+BinPackArguments: false
+BinPackParameters: false
 BraceWrapping:
   AfterCaseLabel:  false
   AfterClass:      false
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ find_package(Qt5Core REQUIRED)
 find_package(Qt5Widgets REQUIRED)
 find_package(Qt5OpenGL REQUIRED)
 # Note: Bluefox SDK is not compatible with some components of OpenCV. Take care when enabling more.
-find_package(OpenCV REQUIRED COMPONENTS core imgproc imgcodecs videoio)
+find_package(OpenCV REQUIRED COMPONENTS core imgproc imgcodecs videoio calib3d)
 
 include(src/shared/CMakeLists.txt.inc)
 
@@ -142,6 +142,7 @@ set (SRCS ${SRCS}
 
     src/app/gui/maskwidget.cpp
 	src/app/gui/automatedcolorcalibwidget.cpp
+	src/app/gui/camera_intrinsic_calib_widget.cpp
 	src/app/gui/cameracalibwidget.cpp
 	src/app/gui/colorpicker.cpp
 	src/app/gui/glLUTwidget.cpp
@@ -155,6 +156,7 @@ set (SRCS ${SRCS}
 
 	src/app/plugins/plugin_mask.cpp
 	src/app/plugins/plugin_cameracalib.cpp
+	src/app/plugins/plugin_camera_intrinsic_calib.cpp
 	src/app/plugins/plugin_colorcalib.cpp
 	src/app/plugins/plugin_colorthreshold.cpp
 	src/app/plugins/plugin_detect_balls.cpp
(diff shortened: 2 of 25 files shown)


g3force:
This PR is ready to be reviewed and tested.

The calibration worked well for me with the FLIR SSL cameras. Compared to the old calibration procedure, distortion and principal point are much more reasonable given enough checkerboard samples.

I did not notice an increase in processing time, but I did not have a field with many robots yet. The imaga2field and field2image functions are slightly slower with the OpenCV method, but this should not be noticeable.

tobiasheineken:
This PR crashes on our PC with 
`(vision:8215) Gtk-ERROR **: 09:42:00.671: GTK+ 2.x symbols detected. Using GTK+ 2.x and GTK+ 3 in the same process is not supported`

gdb tells us 
```
Thread 1 "vision" received signal SIGTRAP, Trace/breakpoint trap.
(gdb) bt
#0  0x00007ffff1e59ea1 in () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#1 0x00007ffff1e5af3c in g_log_default_handler () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#2 0x00007fff1e5b1cd in g_logv () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
...
#14 0x00007ffff6fa75af in QApplication::QApplication(int&, char**, int) () at /usr/lib/x86_64-linux-gnu/libQtGui.so.4
#15 0x00005555556e5284 in main(int, char**) (argc=1, argc=0x7ffffffffe0c08) at /home/robotics/ssl-vision-chess/src/app/main.cpp:85
```

master does not have this crash on startup.

g3force:
can you provide some details on your system like the OS?

tobiasheineken:
Sure. It's a "normal" Xubuntu 18.04:

```
$ uname -a
Linux robolab 4.15.0-88-generic #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
$lsb_release -a
No LSB modules are available
Distributor ID: Ubuntu
Description: Ubuntu 18.04.3 LTS
Release: 18.04
Codename: bionic
$

```

g3force:
Have you run the install script? https://github.com/RoboCup-SSL/ssl-vision/blob/master/InstallPackagesUbuntu.sh

And do you have a self-compiled opencv?

tobiasheineken:
```
$ opencv_version
3.2.0
$ sudo apt install libopencv-dev
libopencv-dev ist schon die neueste Version (3.2.0+dfsg-4ubuntu0.1)
$ sudo apt install libv4l-0
libv4l-0 ist schon die neueste Version (1.14.2-1)

```

Sorry for German printouts. Our opencv is (probably) not self compiled, and while I didn't run this installer script today, we most certainly did so a few months ago and the packages seem to be correct.

g3force:
I can reproduce the error message on a SSL NUC when building with QT4. Building with QT5 seems to fix the issue. You can switch by running:
```
cmake -DUSE_QT5=true build/
make
```
It might be time to make QT5 the default...

tobiasheineken:
This allows me to start the vision. Thanks @g3force. I'll test the PR later.

Any clue why this PR triggers a conflict with QT4?

g3force:
It might be the additional opencv module 'calib3d' that was introduced in this PR.

rhololkeolke:
Are there any other blockers to merging this in once the current conflicts are fixed?

tobiasheineken:
@rhololkeolke I used to test this in our lab (Feb 2020) and it did not work at all. It's reasonable I did something wrong, but I never made it work. Nicolai has published a lot of commits since then, but I cannot test it due to COVID restrictions right now.

g3force:
I improved the capturing of the chessboard samples since then, but it still needs more testing.

The main issue though is, that I'm not yet satisfied with the quality of the result. The overall calibration is not as good as with the old implementation and I could not yet find out the source for this.
I've tried other optimizers, but that did not help. I rather expect an issue in the transformation between image to field and vice versa.

g3force:
This branch is ready for another test round. I've added some notes on the OpenCV calibration to the Wiki: https://github.com/RoboCup-SSL/ssl-vision/wiki/camera-calibration#calibration-with-opencv-model

PR #164, opened 2020-01-16
Remove vertical splitter which has no lower content

g3force:
In the past, there where only two cameras and they
were arranged above each other.
Now we have tabs for each camera/thread and the
lower part is just empty and wastes space.
This commit removes this empty space such that
the camera image and the control widgets expand
to the bottom.

It looks like this now:

image


rhololkeolke:
LGTM

PR #167, opened 2020-02-25
Initialize mask with ones to detect bot pattern images correctly

g3force:
With #158 a image mask has been introduced for thresholding.

The thresholding is also used to extract the blobs from the standard robot pattern (which is an image), in which case a default mask is generated.

Unfortunately, the mask is generated all zero, instead of all one, so the standard robot pattern can not be read correctly.

This PR initializes the mask with ones instead. This fixed the issue for me.


tobiasheineken:
Works for me, thank you.

PR #176, opened 2020-08-10
Feature/image size

### LINKED ISSUE #175: Add Camera Image size to calibration information
Right now, the camera image dimensions are not a part of the SSL_GeometryCameraCalibration. I would like to request that this becomes the case, as such it is then easier to reconstruct if a given ball is within's a camera's view / projects onto the images of the camera . As far as I am aware, it should be relatively simple to obtain this information and store it in the relevant classes.

rolfvdhulst:
Adds the image sizes to the geometryCalibrationData of each camera, allowing to better reconstruct if objects should be visible or not.
fixes #175

### DIFF
--- a/src/client/main.cpp
+++ b/src/client/main.cpp
@@ -138,8 +138,8 @@ int main(int argc, char *argv[])
                     printf("  -tx=%.2f\n",calib.tx());
                     printf("  -ty=%.2f\n",calib.ty());
                     printf("  -tz=%.2f\n",calib.tz());
-                    printf( " -pixel_image_width=%d\n",calib.pixel_image_width());
-                    printf(" -pixel_image_height=%d\n",calib.pixel_image_height());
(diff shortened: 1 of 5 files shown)


g3force:
I've tested it and it looks good to me.

Could also add the new fields to the cli client? (src/client/main.cpp)

rolfvdhulst:
I added the new fields to the CLI client. Thanks for pointing that out :+1:

src/client/main.cpp
```diff
@@ -138,6 +138,8 @@ int main(int argc, char *argv[])
                     printf("  -tx=%.2f\n",calib.tx());
                     printf("  -ty=%.2f\n",calib.ty());
                     printf("  -tz=%.2f\n",calib.tz());
+                    printf( " -pixel_image_width=%d\n",calib.pixel_image_width());
```
g3force, line 141: The formatting and indentation is wrong here and in the next line.

Issue #173, opened 2020-06-05
Basler camera timing errors


rolfvdhulst:
When we (RoboTeam Twente) were testing our vision filters, we noted that often the Basler camera's produced a lot of consistent timing errors. We will probably submit a patch later to fix/amend these issues. I don't know how the other camera's were implemented, but it may be noted that likely doing a strategy along the lines of 'grab latest image' from a camera can be problematic if the image is not timestamped.
I added two graphs here that show the typical raw velocity data of a robot kicking the ball of the basler camera (top) vs a kick from the 2019 robocup logs (bottom). You can clearly see the structure of the noise of the basler camera is periodic. 
![BaslerCameraKick](https://user-images.githubusercontent.com/37153373/83893770-a96df780-a750-11ea-8ecf-feff33463a48.png)
![Robocup2019kick](https://user-images.githubusercontent.com/37153373/83893772-aa068e00-a750-11ea-9272-3c02fbf616f3.png)

g3force:
After a quick look into the source code, it looks like at least basler, spinnaker (current SSL cams) and dc1394 (firewire) all do the same:
```
  timeval tv{};
  gettimeofday(&tv, nullptr);
  result.setTime((double)tv.tv_sec + tv.tv_usec*(1.0E-6));
```
For Spinnaker, there is a timestamp available, but it would need to be synchronized manually to real clock, because ssl-vision is using unix timestamps and assumes synchronized system clocks among all vision computers.

g3force:
One difference that I see is, the basler capture module retrieves the timestamp before the image, the other two modules retrieve the timestamp afterwards, which sounds more reasonable to me.

rolfvdhulst:
Yeah you're right. As far as I can tell from Pylon's documentation it should also be possible to set a frame limit and get the actual capture time from the Camera's System Clock (atleast for the model that we use), but then we may need to synchronize that with the time of the computer running ssl-vision again.

rttwente:
I investigated it, but the moment at which the timestamp is set gives an insignificant error. The difference between getting the timestamp before and after the image is approximately ~10^-5 seconds, which is negligible.

rolfvdhulst:
closed by #177

PR #177, opened 2020-08-18
Fix/basler oscillating timing

rolfvdhulst:
This PR is related to fix issue #173.

The problem turned out to be two-fold.
First of all, we were (accidentally) building in debug. This was not a problem by itself, but it was one of the causes for this issue.

Because in this case, the processing of an image took longer than the interval between two camera frames, ssl-vision would sometimes skip frames from the camera. This, combined with the 'take latest image' strategy and the timestamping being done in SSL-vision, would cause periodic timing errors on the order of ~10 ms. 

Note other camera implementations might face the same issue if their capture rate is higher than the rate SSL-vision can process at and they use a process-latest-image strategy. (I would probably check if this is the case for camera's with large images, such as the ones used in competition).
 
Even though building in Release largely fixed the issue, we still wanted to use the timestamps provided by the camera itself. These were typically more reliable than the ones from ssl-vision. I implemented a moving average filter to measure the average delay between the camera time and the observation time by ssl-vision. Combining this with the time from the camera gives good accuracy on both a higher level (so that more than one camera can be used) and on a lower level (using the exact camera timing from the internal camera clock).

This PR also fixes various segfaults, bugs, exceptions and a deadlock in the basler module, and adds a framerate cap for the camera so that this issue can be prevented in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,8 @@ Then build with the corresponding option:
  * `-DUSE_FLYCAP=true`
  * `-DUSE_V4L=true`
  
-Example: `cd build; cmake -DUSE_SPINNAKER=true ..`. As these are cached cmake option, you only need to run this once and can build with `make` afterwards.
+Example for a release build: `cd build; cmake -DUSE_SPINNAKER=true -DCMAKE_BUILD_TYPE=Release ..`.
+ As these are cached cmake options, you only need to run this once and can build with `make` afterwards.
 
 ### Virtual Splitter cameras
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Then build with the corresponding option:
  * `-DUSE_FLYCAP=true`
  * `-DUSE_V4L=true`
  
-Example for a release build: `cd build; cmake -DUSE_SPINNAKER=true -DCMAKE_BUILD_TYPE=Release ..`.
+Example for a release build: `cd build; cmake -DUSE_SPINNAKER=true ..`.
  As these are cached cmake options, you only need to run this once and can build with `make` afterwards.
 
 ### Virtual Splitter cameras
--- a/src/shared/capture/capture_basler.cpp
+++ b/src/shared/capture/capture_basler.cpp
@@ -110,6 +110,9 @@ bool CaptureBasler::_buildCamera() {
             camera->ChunkSelector.SetValue(Basler_GigECamera::ChunkSelector_Framecounter);
             camera->ChunkEnable.SetValue(true);
             camera->GevTimestampControlReset.Execute(); //Reset the internal time stamp counter of the camera to 0
+            timeval tv;
+            gettimeofday(&tv, 0);
+            initialOffset = tv.tv_sec + (tv.tv_usec / 1000000.0);
         }else{
 		    return false; //Camera does not support accurate timings
 		}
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -53,7 +53,8 @@ Then build with the corresponding option:
  * `-DUSE_FLYCAP=true`
  * `-DUSE_V4L=true`
  
-Example: `cd build; cmake -DUSE_SPINNAKER=true ..`. As these are cached cmake option, you only need to run this once and can build with `make` afterwards.
+Example for a release build: `cd build; cmake -DUSE_SPINNAKER=true -DCMAKE_BUILD_TYPE=Release ..`.
```
g3force, line 56: the build type is also passed to cmake in the Makefile, so specifying it here should not have any effect.

src/shared/capture/capture_basler.cpp
```diff
@@ -101,8 +104,20 @@ bool CaptureBasler::_buildCamera() {
 				Pylon::CTlFactory::GetInstance().CreateDevice(info));
         printf("Opening camera %d...\n", current_id);
 		camera->Open();
-        printf("Setting interpacket delay...\n");
-		camera->GevSCPD.SetValue(600);
+        camera->GammaSelector.SetValue(Basler_GigECamera::GammaSelector_User); //Necessary for interface to work
+        camera->AcquisitionFrameRateEnable.SetValue(true); //Turn on capped framerates
+        //let camera send timestamps and FrameCounts.
+		if(GenApi::IsWritable(camera->ChunkModeActive)){
```
g3force, line 110: mixed tab and space indentation

src/shared/capture/capture_basler.cpp
```diff
@@ -231,9 +247,21 @@ RawImage CaptureBasler::getFrame() {
 		unsigned char* buf = (unsigned char*) malloc(capture.GetImageSize());
 		memcpy(buf, capture.GetBuffer(), capture.GetImageSize());
 		img.setData(buf);
-
 		last_buf = buf;
 
+		if(grab_result->GetPayloadType() == Pylon::PayloadType_ChunkData &&
+		GenApi::IsReadable(grab_result->ChunkTimestamp)){
+		    //const unsigned long int freq =camera->GevTimestampTickFrequency.GetValue(); // This should always be 125 MHz for Basler-ace-1300-75gc
```
g3force, line 254: Can this differ between cameras? Then it might be reasonable to retrieve the value once on startup.

src/shared/capture/capture_basler.cpp
```diff
@@ -317,25 +346,28 @@ void CaptureBasler::resetCamera(unsigned int new_id) {
...
 }
 
-void CaptureBasler::writeParameterValues(VarList* vars) {
-	if (vars != this->settings) {
+void CaptureBasler::writeParameterValues(VarList* varList) {
+	if (varList != this->settings) {
 		return;
 	}
-	bool restart = is_capturing;
+	//bool restart = is_capturing;
```
g3force, line 353: can the code be removed?

README.md
rolfvdhulst: Fixed this. We were not aware, we always ran make in the CMake directory

src/shared/capture/capture_basler.cpp
rolfvdhulst: Good suggestion, I changed this

src/shared/capture/capture_basler.cpp
rolfvdhulst: What are the correct indentation settings for this project? I tried changing it but not sure if this is better now.

g3force: It is not consistent unfortunately... Just make sure it fits around the existing code.

g3force (approved):
LGTM

Do know of any other team that uses Basler camera and who could test it? Otherwise, I'll merge it the next days.

rolfvdhulst:
> LGTM
> 
> Do know of any other team that uses Basler camera and who could test it? Otherwise, I'll merge it the next days.

As far as I am aware, RoboTeam Twente is the only team using the Basler Camera's. I asked in the discord roughly 2 months ago and no one replied.

PR #179, opened 2020-11-28
Use the camera time by syncing it with the system time

g3force:
Some camera APIs provide a camera-internal time for each image. This is much more accurate than using the system time.

The camera-internal time must be converted to unix time, because consumers of the ssl-vision message expect the timestamp to be in unix time. This is especially important when using multiple ssl-vision instances.

### DIFF
--- a/src/shared/capture/capture_basler.cpp
+++ b/src/shared/capture/capture_basler.cpp
@@ -250,7 +250,8 @@ RawImage CaptureBasler::getFrame() {
 
 		if(grab_result->GetPayloadType() == Pylon::PayloadType_ChunkData &&
 		GenApi::IsReadable(grab_result->ChunkTimestamp)){
-                    uint64_t  image_timestamp = grab_result->ChunkTimestamp.GetValue();
+                    double period = 1e9 / camera_frequency;
+                    uint64_t image_timestamp = period * grab_result->ChunkTimestamp.GetValue();
                     timeSync.update(image_timestamp);
                     double time = timeSync.sync(image_timestamp) / 1e9;
 		    img.setTime(time);
--- a/src/shared/capture/capture_basler.cpp
+++ b/src/shared/capture/capture_basler.cpp
@@ -235,9 +235,6 @@ RawImage CaptureBasler::getFrame() {
 			MUTEX_UNLOCK;
 			return img;
 		}
-        timeval tv;
-        gettimeofday(&tv, 0);
-        double computerTime = (double) tv.tv_sec + (tv.tv_usec / 1000000.0);
 
 		Pylon::CPylonImage capture;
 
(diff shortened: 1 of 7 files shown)


g3force:
Tested with the Spinnaker camera on a SSL NUC

Using system time:
image

Using new TimeSync:
image

g3force:
@rolfvdhulst Could you test if it compiles and runs with the Basler module? I do not have the SDK installed.

src/shared/capture/capture_basler.cpp
```diff
@@ -253,13 +250,15 @@ RawImage CaptureBasler::getFrame() {
...
-		    const double period = 1.0 / (double) camera_frequency;// freq should always be 125 MHz for Basler-ace-1300-75gc
-		    long unsigned int timeStamp = grab_result->ChunkTimestamp.GetValue();
-		    double baslerTime = period* (double) timeStamp;
-		    addOffset(computerTime-baslerTime);
-		    img.setTime(getAverageOffset()+baslerTime);
-		}else{
-		    img.setTime(computerTime);
+                    uint64_t  image_timestamp = grab_result->ChunkTimestamp.GetValue();
+                    timeSync.update(image_timestamp);
+                    double time = timeSync.sync(image_timestamp) / 1e9;
```
rolfvdhulst, line 255: The time from the basler camera's is not in Nanoseconds but rather in internal system ticks, which can vary for each camera or even configuration, so you have to get it at startup using camera->GetTimestampTickFrequency.GetValue() as in line 109. The most common value here is 125 MHz, (which is the default in the header). Line 109 also doesn't compile yet on my system. I suggest, adding camera_frequency to the header again.

src/shared/capture/capture_basler.h
```diff
@@ -63,21 +64,14 @@ class CaptureBasler: public QObject, public CaptureInterface {
...
-	bool ignore_capture_failure;
+        TimeSync timeSync;
+        bool ignore_capture_failure;
 	Pylon::CBaslerGigEInstantCamera* camera;
 	Pylon::CBaslerGigEGrabResultPtr grab_result;
 	Pylon::CImageFormatConverter converter;
 	unsigned int current_id;
   	unsigned char* last_buf;
 
-  	int camera_frequency = 125000000;
```
rolfvdhulst, line 73: I suggest keeping this field, see above.

rolfvdhulst (commented):
This is not compiling for the Basler Camera's.

rolfvdhulst:
Do want to say I really appreciate this work though, I think this will help vision quality during matches a lot (at least, we saw notable differences).

g3force:
Thx. I've reverted the removal of the camera frequency.

If it does not work, could you apply your suggested fixes? I think, as a reviewer you should even have push access to the branch :thinking:

rolfvdhulst:
Compiles properly now! We cannot physically test with the Basler Camera's for a while (altleast a few months), so I suggest merging this. In the case of any issues a new PR could be made.

PR #180, opened 2021-01-20
Add ball model configuration to protobuf

g3force:
Most teams use some kind of ball model in their software to track and predict the ball. These models require certain parameters.

One common model for straight kicks is the two phase model, with a sliding and a rolling phase and a switch factor. It is used by ER-Force and TIGERs and their autoRefs.

Instead of doing the configuration for each software (especially the autoRefs) individually, it would be nice to have the parameters at a central place and read it from there.

This PR proposes to add the parameter to the ssl-vision protocol and allow configuration within ssl-vision.

### DIFF
--- a/src/app/plugins/plugin_publishgeometry.cpp
+++ b/src/app/plugins/plugin_publishgeometry.cpp
@@ -56,8 +56,8 @@ string PluginPublishGeometry::getName() {
 
 void PluginPublishGeometry::sendGeometry() {
   SSL_GeometryData geodata;
-  SSL_GeometryFieldSize * gfield = geodata.mutable_field();
-  _field.toProtoBuffer(*gfield);
+  _field.toProtoBuffer(*geodata.mutable_field());
+  _field.toProtoBuffer(*geodata.mutable_models());
   for (unsigned int i = 0; i < params.size(); i++) {
     int camId = params[i]->additional_calibration_information->camera_index->get();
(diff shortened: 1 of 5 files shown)


rhololkeolke:
Do you plan to add a calibration routine for these somewhere? 

I see there are some default values which I assume you took from the TIGERs autoref. But I expect these numbers to vary and I'm not sure how useful it will be without some way of estimating them inside the software.

g3force:
The values vary a lot between different fields, that's why we want to have them at a central place.

We have an mostly-automated mechanism in the TIGERs software, but this involves our robots. You need to really kick and chip the ball, to get realistic data. So I do not see how this would be implemented in ssl-vision.
But as teams usually also need to determine these values, I would hope that the TC can gather the right parameters from one of them for each competition.

rhololkeolke:
That's fine. I just wanted to clarify what the plan was for getting these numbers on a new field. 

The comments on each of the individual fields are good . But I think we should also add a comment on the message that gives the relevant friction model equations. So its clearer how these are combined to make predictions about velocity, acceleration, forces, etc.

Otherwise, we can merge this in.

g3force:
For the two phase model, I have added a reference to the ER-Force TDP that describes the model in more detail.

As both models are not continuous, there is no simple formula. And the protobuf file is probably not the right place for a more complex formula or mathematical description. I would just keep it as it is now, when the descriptions of the parameters are sufficiently clear.

PR #188, opened 2022-01-20
Allow quickly applying geometry settings for both divisions

g3force:


src/shared/util/field_default_constants.h
```diff
@@ -32,18 +32,37 @@ class FieldCircularArc;
...
-    const double kPenaltyAreaWidth = 2400.0;
+    const double kPenaltyAreaDepth = 1800.0;
+    const double kPenaltyAreaWidth = 3600.0;
     const double kCenterCircleRadius = 500.0;
     const double kGoalLineToPenaltyMark = 8000.0;
     const double kBallRadius = 21.5;
     const double kMaxRobotRadius = 90.0;
     const int kNumCamerasTotal = 2;
-    const int kNumCamerasLocal = 2;
+    const int kNumCamerasLocal = 1;
```
lhcavalcanti, line 47: Just to make sure, this means the number of camera per ssl-vision, right? So, we use two computers for vision at Div A field, one for each camera? (I just remember the Div B setup from Sydney was 1 computer and 1 camera)

lhcavalcanti (approved):
Everything looks good! Just made a comment to clarify the setup.

src/shared/util/field_default_constants.h
g3force: correct. Local in this case means, on the current computer/ssl-vision instance. For division A, we use two NUC computers, each having one camera. The parameters influence the generation of the camera calibration control points.

PR #189, opened 2022-03-12
Add the framerate setting for the V4L camera

akito914:
In the previous code, it seemed that the framerate specified in the GUI was not reflected in behavior. 
This pull request adds a framerate argument to the GlobalV4Linstance::startStreaming() method.
I have confirmed that "Logicool c922n" works at 60 frames/sec in MJPEG format.


g3force:
looks good to me

PR #194, opened 2022-11-09
Implemented nonstop recording and writing at limited FPS. Current problem : small stutter when writing.

emielsteerneman:
This PR adds a second recording mode to SSL-Vision. This mode records frames (.png) and accompanying detection frames (.json) nonstop, and immediately writes these to disk. An optional (and enabled by default) FPS limiter will prevent the PC from being flooded with write operations. 

The collection of this data could help develop SSL-Vision. For example, the data could be used to train models for automatic color calibration, as well as adaptive color calibration when lighting conditions change.

Currently, whenever an image is written, there is a slight stutter in that camera its thread. The writing should be done in a different thread, but this has not yet been implemented. Because of this slight stutter, this feature is currently unusable for actual Robocup matches. It is usable for anything outside an official RoboCup match.


emielsteerneman:
Merged the wrong branch...

PR #197, opened 2022-11-10
Remove QT4

emielsteerneman:
Completey removed QT4 from SSL-Vision. QT4 is not supported anymore [since 2015](https://en.wikipedia.org/wiki/Qt_version_history#Qt_4), and has been dropped from Ubuntu 20.04 and 22.04.


g3force:
thx, lgtm :+1:

Issue #198, opened 2022-11-24
build fails with -DUSE_DC1394=false when libdc1394 is not installed


kenshin33:
compilation fails to finish complaining a bout a missing header provided by libdc1394

[this line](https://github.com/RoboCup-SSL/ssl-vision/blob/64f3f29929ccd3f7cb316102454decd624056d06/src/app/capture_thread.h#L24) should be guarded by either DC1394 or  NO_DC1394_CONVERSIONS as it was done  in https://github.com/RoboCup-SSL/ssl-vision/commit/979dc066892153df1e3ab6f46ea3177873352721

g3force:
Interesting. Looks like nobody ever tested this before.

Would you mind creating a PR with your suggested fix?

We could also remove the package from the installation script and only mention the required package in the Readme.

kenshin33:
sure thing . But for the record, I have no idea what this project does or how it works. I just stumbled on this issue while helping someone build it!

kenshin33:
fixed in https://github.com/RoboCup-SSL/ssl-vision/pull/199

PR #199, opened 2022-11-24
guard the inclusion of "capturedc1394v2.h" with DC1394

kenshin33:
if  USE_DC1394 was false,  the inlusion caused the build to fail
apossible fix for #198


g3force:
thx for your contribution :+1:

PR #201, opened 2023-01-15
fix for using V4L-mjpeg

nabeya11:
Once bad jpeg image was got, Vision would terminate.
Then, I modified libjepg error handler.
cf. https://stackoverflow.com/questions/19857766/error-handling-in-libjpeg


g3force:
Looks good, thx for the contribution :+1:

PR #202, opened 2023-01-30
Added default camera ID for Basler

emielsteerneman:
With this, the Basler cameras are properly enumerated 0,1,2,3 instead of 0,0,0,0. Tested it, and it works.


g3force:
Nice to hear that it even works :upside_down_face:

PR #203, opened 2023-02-23
Fix and speed up v4l yuv2rgb conversion

akito914:
Fixed incorrect YUV2RGB conversion. 
And the conversion had large overhead due to per-pixel function calls, so this has been fixed.


g3force:
Sounds good, thx for the contribution :+1:

PR #209, opened 2023-05-10
Fix error on installing test data

ryota2357:
`make install_test_data` is failed because of not exists test-data directory.


g3force:
good catch :+1:

PR #217, opened 2024-05-04
Implement OpenCV Video source

lixfel:
Enables the usage of typical video formats as source for ssl-vision for testing and benchmarking.

The impact of the additional OpenCV module videoio on the Bluefox SDK support has not been tested.

### DIFF
--- a/src/shared/capture/capture_video.cpp
+++ b/src/shared/capture/capture_video.cpp
@@ -39,6 +39,11 @@ RawImage CaptureVideo::getFrame() {
   img.ensure_allocation(ColorFormat::COLOR_RGB8, factor*frame.cols, factor*frame.rows);
   cv::Mat dstImg(img.getHeight(), img.getWidth(), CV_8UC3, img.getData());
 
+  /* ssl-vision expects typical color images scaled up from the raw bayer matrix
+   * (4 sensor fields RG-GB are interpolated into four RGB pixels).
+   * To support downscaled image formats (4 color fields RG-GB into 1 RGB value)
+   * where ssl-vision needs the interpolated resolution to work as usual
+   * this upscaling option has been added. */
   if (v_cap_upscale->getBool()) {
     cv::resize(frame, dstImg, dstImg.size());
     cvtColor(dstImg, dstImg, cv::COLOR_BGR2RGB);
@@ -53,7 +58,9 @@ RawImage CaptureVideo::getFrame() {
   return img;
 }
 
-void CaptureVideo::releaseFrame() {}
+void CaptureVideo::releaseFrame() {
+  // frame and img are not cleared here to prevent unnecessary reallocation with the next frame.
+}
 
 
 bool CaptureVideo::startCapture() {
--- a/src/shared/capture/capture_video.cpp
+++ b/src/shared/capture/capture_video.cpp
@@ -30,7 +30,6 @@ CaptureVideo::CaptureVideo(VarList* settings) : CaptureInterface(settings) {
 
 
 RawImage CaptureVideo::getFrame() {
-  cv::Mat frame;
   if(!capture.read(frame)) {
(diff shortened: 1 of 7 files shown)


src/shared/capture/capture_video.cpp
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  timestamp += 1.0/capture.get(cv::CAP_PROP_FPS);
+  img.setTimeCam(timestamp);
+
+  return img;
+}
+
+void CaptureVideo::releaseFrame() {
+  //img.clear();
```
g3force, line 58: Why is this commented out?

src/shared/capture/captureinterface.cpp
```diff
@@ -46,6 +46,9 @@ void CaptureInterface::readAllParameterValues() {
 }
 
 bool CaptureInterface::copyAndConvertFrame(const RawImage & src, RawImage & target) {
+  if(target.getColorFormat() == COLOR_UNDEFINED)
```
g3force, line 49: Why was this necessary? Maybe add a short comment, explaining it.

src/shared/capture/capture_video.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+
+RawImage CaptureVideo::getFrame() {
+  cv::Mat frame;
+  if(!capture.read(frame)) {
+    std::cout << "End of video stream reached" << std::endl;
+    return img;
+  }
+
+  int factor = v_cap_upscale->getBool() ? 2 : 1;
```
g3force, line 39: Wouldn't it make more sense to configure the scale factor, instead of a flag?

src/shared/capture/captureinterface.cpp
lixfel: COLOR_UNDEFINED has a pixel size of 0, which causes a segmentation fault in the following memcpy during the first frame in capture_thread.cpp#327 (`bool bSuccess = capture->copyAndConvertFrame( pic_raw,d->video);`).

src/shared/capture/capture_video.cpp
lixfel: Other scalings should not be needed, this is just necessary due to the half resolution of the videos captured with the new VisionProcessor.

g3force: Would be good to have some explanation of why this is necessary somewhere. I have not quite get why, yet :upside_down_face:

src/shared/capture/captureinterface.cpp
g3force: I still do not understand why your change fixes the issue. The color format should be applied anyways, except if the image is of size zero, in which case the data is equal to `nullptr`.

src/shared/capture/capture_video.cpp
```diff
@@ -0,0 +1,81 @@
...
+    cvtColor(frame, dstImg, cv::COLOR_BGR2RGB);
+  }
+
+  timestamp += 1.0/capture.get(cv::CAP_PROP_FPS);
+  img.setTimeCam(timestamp);
+
+  return img;
+}
+
+void CaptureVideo::releaseFrame() {}
```
g3force, line 56: Why not clearing the matrices here? At least add a comment with that explanation.

src/shared/capture/capture_video.cpp
lixfel: Typical color cameras do an upscaling during the conversion from the Bayer matrix to the color image. So 4 sensor fields (RG - GB) are interpolated into 4 RGB pixels. To reduce necessary bandwidth and improve image processing time my new vision processor does an "downsampling" (only affects the green color channel) instead (RG - GB becomes 1 RGB pixel). To use these "half" resolution videos with the same geometry and blob resolution as the regular ssl-vision is used to this upscaling option is necessary.

PR #238, opened 2026-06-11
Update ER-Force robot height to 148mm

chrismit3s:
Our robot height changed a little (2mm), this updates the vision config to match it.

Were can I adjust the robot height for the vision processor?


g3force:
@lixfel FYI

